Issue #2, opened 2020-10-30
Set logs to match local TZ of device


briankelly:
currently the mobile logs are presenting time in UTC. i suggest using local TZ instead, or provide a way for user to configure this

PR #8, opened 2020-12-16
Log timezone 

micha3lbrown:
The default log timestamp is in UTC, this commit allows the user to
convert the log timestamps to their own timezone.

When approaching this problem I originally considered allowing the user
to set any timezone offset of their choice. This would require some
alignment between the device and client that was feeling cumbersome.

In the end we decided to keep things relatively simple and just allow a
simple toggle to convert log entries upon reading the logfile on the
device.

The default log timestamp emitted by the client remains UTC/Zulu

Advances #2


micha3lbrown:
I have confirmed functionality works as expected for Android, additional work is required to confirm iOS.

Issue #3, opened 2020-10-30
option to enable text wrapping


briankelly:
for logs and configs and anywhere else

personally, i prefer this to be optional, and to keep the default as it is currently presented (nowrap)

jasikpark:
Fixed by https://github.com/DefinedNet/mobile_nebula/pull/10, added a global option for toggling wrapping the logs. Other text is already wrapping by default.

<img width="351" alt="Image" src="https://github.com/user-attachments/assets/47e97a05-f9e3-403f-9b8c-f3412af11478" />

PR #10, opened 2021-01-20
Add advanced config option for logWrap

micha3lbrown:
Add Advanced config option to change display of log formatting.
`logWrap` only impacts logs and does not enable config wrapping at this
time.

Advances #3


android/app/src/main/kotlin/net/defined/mobile_nebula/Sites.kt
```diff
@@ -251,6 +251,7 @@ class IncomingSite(
     val cipher: String,
     val sortKey: Int?,
     var logVerbosity: String?,
+    val logWrap: Boolean?,
```
nbrownus, line 254: We can undo this and it's friends now that it isn't a side based config

lib/models/Site.dart
```diff
@@ -11,7 +11,8 @@ import 'StaticHosts.dart';
 var uuid = Uuid();
 
 class Site {
-  static const platform = MethodChannel('net.defined.mobileNebula/NebulaVpnService');
+  static const platform =
```
nbrownus, line 14: Looks like your IDE may have taken a large hammer and adjusted much whitespace. I don't find a lot of value in it personally. Happy to discuss further if others have strong feelings.

I'll continue after we resolve these first two

android/app/src/main/kotlin/net/defined/mobile_nebula/Sites.kt
micha3lbrown: Fixed! ✅

lib/models/Site.dart
micha3lbrown: I've pulled out most of the flutter formatting which should make salient changes easy to identify. 👍

lib/screens/SiteLogsScreen.dart
```diff
@@ -25,6 +26,7 @@ class _SiteLogsScreenState extends State<SiteLogsScreen> {
   ScrollController controller = ScrollController();
   RefreshController refreshController = RefreshController(initialRefresh: false);
 
+  var settings = Settings();
```
nbrownus, line 29: This should be a member of `_SiteLogsScreenState` to avoid any global state

nbrownus: And here is proof that I do not know how to read

nbrownus (approved):

PR #11, opened 2021-02-01
Fix 'cannot create regular file` error on initial build

mediocregopher:
If you've never built the project before the `android/app/src/main/libs`
directory doesn't exist, and so `./gen-artifacts.sh` errors out when run
in that case. This commit fixes `./gen-artifacts.sh` so that it makes
that directory before trying to move a file there. It also removes an
unnecessary `rm -rf` command.


nbrownus:
Thanks for the PR! This change was included in #25. Now that the CLA system is set up we can merge outside contributions going forward.

PR #13, opened 2021-02-05
Add logLocalTZ to app settings

micha3lbrown:
The default log timestamp is in UTC, this commit allows the user to
convert the log timestamps to their own timezone.

When approaching this problem I originally considered allowing the user
to set any timezone offset of their choice. This would require some
alignment between the device and client that was feeling cumbersome.

In the end we decided to keep things relatively simple and just allow a
simple toggle to convert log entries upon reading the logfile on the
device.

The default log timestamp emitted by the client remains UTC/Zulu

Advances #2
Closes #8

### DIFF
--- a/lib/screens/SiteLogsScreen.dart
+++ b/lib/screens/SiteLogsScreen.dart
@@ -103,8 +103,11 @@ class _SiteLogsScreenState extends State<SiteLogsScreen> {
   loadLogs() async {
     var file = File(widget.site.logFile);
     try {
-      final v = await file.readAsString();
+      String v = await file.readAsString();
 
+      if (settings.logLocalTZ) {
+        v = convertToLocalTZ(v);
+      }
       setState(() {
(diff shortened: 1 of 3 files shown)


lib/screens/SiteLogsScreen.dart
```diff
@@ -103,8 +103,11 @@ class _SiteLogsScreenState extends State<SiteLogsScreen> {
   loadLogs() async {
     var file = File(widget.site.logFile);
     try {
-      final v = await file.readAsString();
+      String v = await file.readAsString();
```
micha3lbrown, line 106: @nbrownus this was one change I was unsure about. I believe I had to convert the variable type from `final` to `String` to allow modifications. Let me know if this looks incorrect to you.

micha3lbrown:
IMG_FF3216914A7C-1

Here is how the config section looks on iOS.

PR #24, opened 2021-04-22
IPV6 support

nbrownus:
- Binds against nebula core @ `1deb5d9` to give us ipv6 support
- Allows for ipv6 addresses in the host info screen and while configuring static host maps
- Detects network changes and triggers a lighthouse update (which may cascade to causing hole punching, from the other end of a tunnel, improving tunnel performance when roaming)
- Attempts to sleep/wake on Android, non lighthouse tunnels are torn down on sleep. We may want to back this out after testing.


android/app/build.gradle
```diff
@@ -32,7 +32,7 @@ if (keystorePropertiesFile.exists()) {
 }
 
 android {
-    compileSdkVersion 28
+    compileSdkVersion 29
```
JohnMaguire, line 35: Does this remove support for 20% of Android users? https://gs.statcounter.com/os-version-market-share/android

JohnMaguire: Ah, target vs. minimum. I need to read about the differences. Ignore me.

brad-defined (approved):

lib/models/IPAndPort.dart
```diff
@@ -14,12 +20,9 @@ class IPAndPort {
...
   IPAndPort.fromString(String val) {
-    final parts = val.split(':');
-    if (parts.length != 2) {
-      throw 'Invalid IPAndPort string';
-    }
-
-    ip = parts[0];
-    port = int.parse(parts[1]);
+    //TODO: Uri.parse is as close as I could get to parsing both ipv4 and v6 addresses with a port without bringing a whole mess of code into here
+    final uri = Uri.parse("ugh://$val");
```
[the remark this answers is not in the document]
nbrownus: Gah, I did it with `://` and then just went with `ugh://` when that failed as a test. Ended up kinda liking it :)

PR #26, opened 2021-04-23
Upgrade to flutter 2

nbrownus:
This PR currently does not address any new functionality we could/should be using, just gets the build working.


lib/components/config/ConfigButtonItem.dart
```diff
@@ -21,21 +20,5 @@ class ConfigButtonItem extends StatelessWidget {
           constraints: BoxConstraints(minHeight: Utils.minInteractiveSize, minWidth: double.infinity),
           child: Center(child: content),
         ));
-
```
JohnMaguire, line 24: This change seems kind of major compared to the other API fixes in this PR. Can you explain why we are removing this?

lib/screens/siteConfig/CertificateScreen.dart
```diff
@@ -219,34 +219,22 @@ class _CertificateScreenState extends State<CertificateScreen> {
...
                     return;
                   }
-                } catch (err) {
-                  print('HEY $err');
-                }
 
-                var content = "";
-                try {
-                  content = file.readAsStringSync();
+                  _addCertEntry(content, (err) {
```
JohnMaguire, line 228: It seems potentially unnecessary to move this into the try`` - errors should invoke the callback here? I don't see any harm though.

lib/screens/siteConfig/CertificateScreen.dart
```diff
@@ -219,34 +219,22 @@ class _CertificateScreenState extends State<CertificateScreen> {
...
+                    if (err != null) {
+                      Utils.popError(context, 'Error loading certificate file', err);
```
JohnMaguire, line 230: Should this return? Previously this was the last line of code in this function - now the catch comes after it. I could see someone adding code following the catch, and not realizing that this doesn't return in the case of an error.

JohnMaguire: Ignore me - this is occurring as part of a callback. The order confused me a bit.

JohnMaguire (approved):

lib/components/config/ConfigButtonItem.dart
nbrownus: It's unreachable code I left around for whatever reason

PR #28, opened 2021-04-26
Rework ios ipc

nbrownus:
I set out to fix the network gateway detection to not be so verbose and got frustrated with the ios ipc lossy behavior.

This change does two main things.
1. Tries to rebind less by understanding what changed and not rebinding when nothing changed. This is better than before but we can likely improve further.
2. Replaced `MMWormhole`, a file based IPC mechanism with the `NETunnelProvider.handleAppMessage`, these calls can only be triggered by the UI which matches how everything but starting the VPN works before.

The biggest change is the final point, we want to receive startup errors but ios does not give us access to them and we have no way to call the UI directly from the vpn service. So, when the UI asks the system to start we send an option that our vpn code understands as "wait for another start command", this command will come from the UI and has a callback we can use to return success/failure. The actual start command is actually triggered by the system when it notifies us that our vpn service has transitioned from `disconnected` to `connecting`

### DIFF
--- a/ios/NebulaNetworkExtension/PacketTunnelProvider.swift
+++ b/ios/NebulaNetworkExtension/PacketTunnelProvider.swift
@@ -1,6 +1,7 @@
 import NetworkExtension
 import MobileNebula
 import os.log
+import SwiftyJSON
 
 class PacketTunnelProvider: NEPacketTunnelProvider {
     private var networkMonitor: NWPathMonitor?
@@ -15,7 +16,7 @@ class PacketTunnelProvider: NEPacketTunnelProvider {
     // This is the system completionHandler, only set when we expect the UI to ask us to actually start so that errors can flow back to the UI
     private var startCompleter: ((Error?) -> Void)?
     
-    private func log(_ message: StaticString, _ args: CVarArg...) {
+    private func log(_ message: StaticString, _ args: Any...) {
         os_log(message, log: _log, args)
     }
     
@@ -194,7 +195,7 @@ class PacketTunnelProvider: NEPacketTunnelProvider {
                         
                     } else {
                         // Error response has
-                        completionHandler!(try? JSONEncoder().encode(IPCResponse.init(type: .error, message: .string(error!.localizedDescription))))
+                        completionHandler!(try? JSONEncoder().encode(IPCResponse.init(type: .error, message: JSON(error!.localizedDescription))))
                     }
                 }
             }
@@ -220,7 +221,7 @@ class PacketTunnelProvider: NEPacketTunnelProvider {
         }
         
         if (error != nil) {
-            completionHandler!(try? JSONEncoder().encode(IPCResponse.init(type: .error, message: JSON.string(error?.localizedDescription ?? "Unknown error"))))
+            completionHandler!(try? JSONEncoder().encode(IPCResponse.init(type: .error, message: JSON(error?.localizedDescription ?? "Unknown error"))))
         } else {
             completionHandler!(try? JSONEncoder().encode(IPCResponse.init(type: .success, message: data)))
         }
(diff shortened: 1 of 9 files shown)


ios/NebulaNetworkExtension/PacketTunnelProvider.swift
```diff
@@ -132,59 +135,118 @@ class PacketTunnelProvider: NEPacketTunnelProvider {
...
+            if routeDescription != "" {
+                nebula?.rebind("network change to: \(routeDescription); from: \(cachedRouteDescription ?? "none")")
+            }
+            cachedRouteDescription = routeDescription
+        }
+    }
+    
+    private func collectAddresses(endpoints: [Network.NWEndpoint]) -> String {
+        var str: [String] = []
+        endpoints.forEach{ endpoint in
```
JohnMaguire, line 150: Is order guaranteed? If not, the resulting string may not match even though the addresses are all the same.

On that note - is it necessary to convert this to a string, or can we just compare the list values?

ios/NebulaNetworkExtension/PacketTunnelProvider.swift
```diff
@@ -132,59 +135,118 @@ class PacketTunnelProvider: NEPacketTunnelProvider {
...
+                        // Error response has
+                        completionHandler!(try? JSONEncoder().encode(IPCResponse.init(type: .error, message: .string(error!.localizedDescription))))
+                    }
+                }
+            }
+            return
+        }
+        
+        if nebula == nil {
+            // Respond with an empty success message in the event a command comes in before we've truly started
```
JohnMaguire, line 205: Why success instead of failure? I'm not totally sure what conditions this could occur, but a message seems to imply an action should be taken, and we are not taking that action?

ios/NebulaNetworkExtension/Site.swift
```diff
@@ -3,56 +3,124 @@ import MobileNebula
...
+            let encoder = JSONEncoder()
+            encoder.outputFormatting = [.prettyPrinted]
+            return try! String(data: encoder.encode(self), encoding: .utf8)!
+        }
     }
+}
+
+class IPCResponse: Codable {
+    var type: IPCResponseType
+    //TODO: change message to data?
```
JohnMaguire, line 104: Is this just referring to the name of the var?

ios/Runner/AppDelegate.swift
```diff
@@ -143,12 +138,15 @@ func MissingArgumentError(message: String, details: Any?) -> FlutterError {
     func startSite(call: FlutterMethodCall, result: @escaping FlutterResult) {
         guard let args = call.arguments as? Dictionary<String, String> else { return result(NoArgumentsError()) }
         guard let id = args["id"] else { return result(MissingArgumentError(message: "id is a required argument")) }
+        
 #if targetEnvironment(simulator)
         let updater = self.sites?.getUpdater(id: id)
+        activeSite = updater?.getSite()
```
JohnMaguire, line 144: Is this being used for something?

ios/NebulaNetworkExtension/Site.swift
```diff
@@ -151,18 +219,17 @@ struct Site: Codable {
     var cipher: String
     var sortKey: Int
     var logVerbosity: String
-    var connected: Bool?
+    var connected: Bool? //TODO: active is a better name
```
JohnMaguire, line 222: Does this actually need to be optionally null, or can this default false? I was surprised to see the `??` for this elsewhere in the code.

ios/NebulaNetworkExtension/PacketTunnelProvider.swift
nbrownus: Just below is a sort on the resulting string array, they are stored as a full string because this is passed to nebula core as a reason for the rebind and shows up in your logs.

ios/NebulaNetworkExtension/PacketTunnelProvider.swift
nbrownus: The primary purpose right now is to avoid a race where the UI starts issuing commands before core is actually booted, having an error return might make those races more apparent but at the expense of spitting errors at end users and they can't do anything about it.

ios/NebulaNetworkExtension/Site.swift
nbrownus: Yes

ios/Runner/AppDelegate.swift
nbrownus: This is specific to enabling the simulator, which lacks network extension support. Only used in dev to avoid having to pick up another device.

ios/NebulaNetworkExtension/Site.swift
nbrownus: Pedantically, we don't know the state of this until we query the manager for its state and there are ways to hydrate this class without having access to a manger.

ios/NebulaNetworkExtension/PacketTunnelProvider.swift
JohnMaguire: @nbrownus Makes sense - can we log anything, or not easily from here?

JohnMaguire (approved):

ios/NebulaNetworkExtension/PacketTunnelProvider.swift
nbrownus: We can and I added a log statement, nice catch

ios/Runner/AppDelegate.swift
nbrownus: And deleted, nice catch!

lib/screens/SiteDetailScreen.dart
```diff
@@ -89,7 +89,7 @@ class _SiteDetailScreenState extends State<SiteDetailScreen> {
         }),
         refreshController: refreshController,
         onRefresh: () async {
-          if (site.connected) {
+          if (site.connected && site.status == "Connected") {
```
JohnMaguire, line 92: nit: Maybe a constant? Meh.

JohnMaguire (approved):

lib/screens/SiteDetailScreen.dart
nbrownus: Oh man, yes. But also more!

`site.connected` really means `site.active` so I wanna rename that
`site.connected` should then become a boolean meaning "i am actually truly connected now" or replaced with an int status to check against, the computer counterpart to `site.status`
and finally `site.status` stays as the human string.

PR #29, opened 2021-04-29
Kotlin active site race on boot fix

nbrownus:
@JohnMaguire noticed that the UI can get out of sync with the system state and we tracked it down to a race condition between asking the active vpn process what is running and the UI registering event channels to hear about those updates on UI boot.

This fixes that.

It also updates our `go.mod` to point at the most recent nebula master sha.


JohnMaguire (approved):
LGTM!

PR #43, opened 2021-06-08
Add a changelog

nbrownus:


ios/Runner.xcodeproj/project.pbxproj
```diff
@@ -819,7 +819,7 @@
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				CLANG_ENABLE_MODULES = YES;
 				CODE_SIGN_ENTITLEMENTS = Runner/Runner.entitlements;
-				CURRENT_PROJECT_VERSION = 11;
+				CURRENT_PROJECT_VERSION = 1;
```
brad-defined, line 822: Why is the project version dropping from 11 to 1?

nbrownus: This is actually the build number, version would be `0.0.41-1`

brad-defined (approved):

PR #47, opened 2021-09-15
Get ready for ios 15

nbrownus:


ios/NebulaNetworkExtension/PacketTunnelProvider.swift
```diff
@@ -56,19 +55,11 @@ class PacketTunnelProvider: NEPacketTunnelProvider {
...
-
-        var ifnameSize = socklen_t(IFNAMSIZ)
-        let ifnamePtr = UnsafeMutablePointer<CChar>.allocate(capacity: Int(ifnameSize))
-        ifnamePtr.initialize(repeating: 0, count: Int(ifnameSize))
-        if getsockopt(fileDescriptor, 2 /* SYSPROTO_CONTROL */, 2 /* UTUN_OPT_IFNAME */, ifnamePtr, &ifnameSize) == 0 {
-            self.ifname = String(cString: ifnamePtr)
+        
+        let fileDescriptor = tunnelFileDescriptor
+        if fileDescriptor == nil {
+            return completionHandler("Unable to locate the tun file descriptor")
```
JohnMaguire, line 61: I see you added a bang down below. That bang implies that we could unwrap a nil and throw. That's not possible because of the nil check up here, but it's not readily obvious when reading the lower part of this function body.

Did you consider an approach like this, which would alleviate the need for the bang later on?

```
        if tunnelFileDescriptor == nil {
            return completionHandler("Unable to locate the tun file descriptor")
        }

        let fileDescriptor = tunnelFileDescriptor!
```

JohnMaguire (approved):
This sure looks like code to me.

JohnMaguire (approved):

PR #57, opened 2022-01-19
Remove special keyboard handling for text inputs

nbrownus:


lib/components/SpecialTextField.dart
```diff
@@ -84,121 +82,43 @@ class _SpecialTextFieldState extends State<SpecialTextField> {
 
   @override
   Widget build(BuildContext context) {
-    return RawKeyboardListener(
-        focusNode: _focusNode,
```
JohnMaguire, line 88: The `_focusNode` class member is still declared - can we remove it?

JohnMaguire (approved):

PR #71, opened 2022-09-30
Automated build and release

nbrownus:
Whenever a new tag match `v[0-9]+.[0-9]+.[0-9]*` is pushed a new build will start and the resulting Android and iOS artifacts will be published to internal testing tracks. A draft release will also be created in github, we should publish the release if we promote either build to production.


.github/workflows/release.yml
```diff
@@ -0,0 +1,152 @@
+name: Create release and upload to Apple and Google
+on:
+  push:
+    tags:
+      # Only builds for tags with a meaningless build number suffix: v1.0.0-1
+      - 'v[0-9]+.[0-9]+.[0-9]*-.*'
```
JohnMaguire, line 6: I think the final two `*` should be `+` still. At minimum the first one.

.github/workflows/release.yml
```diff
@@ -0,0 +1,152 @@
...
+          TOKEN: ${{ secrets.MACHINE_USER_PAT }}
+        run: git config --global url."https://defined-machine:${TOKEN}@github.com".insteadOf "https://github.com"
+
+      - name: Install the appstore connect key material
+        env:
+          AC_API_KEY_SECRET_BASE64: ${{ secrets.AC_API_KEY_SECRET_BASE64 }}
+        run: |
+          AC_API_KEY_SECRET_PATH="$RUNNER_TEMP/key.p8"
+          echo "APP_STORE_CONNECT_API_KEY_KEY_FILEPATH=$AC_API_KEY_SECRET_PATH" >> $GITHUB_ENV
+          echo -n "$AC_API_KEY_SECRET_BASE64" | base64 --decode --output "$AC_API_KEY_SECRET_PATH"
```
JohnMaguire, line 51: Does this work? GNU base64 doesn't support `--output` AFAIK

.github/workflows/release.yml
```diff
@@ -0,0 +1,152 @@
...
+      - name: Publish to Android internal track
+        run: |
+          cd android
+          fastlane release
+
+      - name: Create Release
+        id: create_release
+        uses: actions/create-release@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
brad-defined, line 126: I don't see this secret in the repo

JohnMaguire (dismissed):
Can you do it again, but without pulling in Ruby?

.github/workflows/release.yml
nbrownus: This one is provided by github on every run

.github/workflows/release.yml
nbrownus: Works on a mac :D

brad-defined (approved):

PR #90, opened 2022-11-17
Ignore a nil message, nebula hasn't booted yet

nbrownus:
Alternatively we could handle `nebula = nil` in each RPC and respond with an appropriate empty message body.


jasikpark (commented):
lgtm

JohnMaguire:
> Alternatively we could handle `nebula = nil` in each RPC and respond with an appropriate empty message body.

I have a slight preference for this approach.

JohnMaguire (approved):
I'd rather not special case these, but it sounds like this fixes a bug.

PR #96, opened 2022-11-21
Sometimes the list of remotes can be null

nbrownus:


jasikpark (approved):
lgtm; this sounds like it's what I was hitting

JohnMaguire (approved):

PR #98, opened 2022-11-21
Make user agent consistent on iOS and Android

JohnMaguire:
Instead of Nebula/, Nebula-DEBUG/, and NebulaNetworkExtension/, ensure that the user agent is always MobileNebula/.


jasikpark (approved):
lgtm

nbrownus (approved):

PR #18, opened 2021-03-13
Add ability to add DNS resolvers to a site

mediocregopher:
For most changes I simply copied whatever was being done for
`unsafeRoutes` and fitted it to work for dns resolvers.

TODO This does not currently work for iOS, only android, as I only have
an android device to test with.

Fixes #9

PR #104, opened 2022-12-28
Add ability to add DNS resolvers to a site

0ff:
This PR builds upon https://github.com/DefinedNet/mobile_nebula/pull/18 and updates it to the latest mobile_nebula version.

It **also adds** iOS Support, so that this can finally be used on both platforms.

### DIFF
--- a/android/app/src/main/kotlin/net/defined/mobile_nebula/NebulaVpnService.kt
+++ b/android/app/src/main/kotlin/net/defined/mobile_nebula/NebulaVpnService.kt
@@ -121,6 +121,16 @@ class NebulaVpnService : VpnService() {
             builder.addRoute(unsafeIPNet.network, unsafeIPNet.maskSize.toInt())
         }
 
+        // Add our dns resolvers
+        site!!.dnsResolvers.forEach { dnsResolver ->
+            builder.addDnsServer(dnsResolver)
+        }
+
+        val cm = getSystemService(Context.CONNECTIVITY_SERVICE) as ConnectivityManager
+        cm.allNetworks.forEach { network ->
+            cm.getLinkProperties(network).dnsServers.forEach { builder.addDnsServer(it) }
+        }
+
         try {
             vpnInterface = builder.establish()
             nebula = mobileNebula.MobileNebula.newNebula(site!!.config, site!!.getKey(this), site!!.logFile, vpnInterface!!.detachFd().toLong())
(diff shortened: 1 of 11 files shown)


numinit (approved):
Very nice!

android/app/src/main/kotlin/net/defined/mobile_nebula/NebulaVpnService.kt
```diff
@@ -121,6 +121,16 @@ class NebulaVpnService : VpnService() {
...
         }
 
+        // Add our dns resolvers
+        site!!.dnsResolvers.forEach { dnsResolver ->
+            builder.addDnsServer(dnsResolver)
+        }
+
+        val cm = getSystemService(Context.CONNECTIVITY_SERVICE) as ConnectivityManager
+        cm.allNetworks.forEach { network ->
+            cm.getLinkProperties(network).dnsServers.forEach { builder.addDnsServer(it) }
```
numinit, line 131: ` cm.getLinkProperties(network)?.dnsServers?.forEach { builder.addDnsServer(it) }`

Would not compile otherwise.

numinit (changes requested):

android/app/src/main/kotlin/net/defined/mobile_nebula/NebulaVpnService.kt
numinit: Come to think of it, lines 129 to 132 should be removed entirely, as they cause ta DNS leak on Android when tested with https://mullvad.net/ (but no others, interestingly). Removing these lines fixes the DNS leak. They are also based on a false assumption that the device will not roam between networks.

gpaterno:
Any way to put this forward? Or, at least, some solution addressing this issue?THanks

0ff:
I don't have an android device to check this on and there seems to be no interest at all from the maintainer to get this merged, so I fear you'd have to take it into your own hands.

I'd love to see this going upstream, as for now I have to maintain my own copy of nebula to use the feature, so if I can do anything to get it merged, let me know.

numinit:
I've confirmed it works fine on my Android device.

numinit:
That being said, would you mind implementing this comment? It definitely causes a DNS leak otherwise.

https://github.com/DefinedNet/mobile_nebula/pull/104/files/97dca45ac2fe0568ddb541cbc3d1808f3ba563d3#r1059683727

LennyPenny:
Maybe you could apply for this job and get the PR merged that way ;) https://www.defined.net/jobs/#senior-software-engineer---mobile

PR #114, opened 2023-04-13
Allow IP address entry on non-US locales

### LINKED ISSUE #105: Nebula IP address for static host cannot be input on iPhone
On iPhone the field for the Nebula IP address in "New Statix Host" opens the wrong input keyboard: only digits 0 through 9 and the comma are there.

The public IP field a bit further down opens the correct full keyboard.

The issue does not exist on iPad.

JohnMaguire:
Fixes #105.

Before (image taken from issue above, but reproduced prior to fix):

image

After: 

Screenshot 2023-04-13 at 4 32 44 PM

### DIFF
--- a/lib/components/IPField.dart
+++ b/lib/components/IPField.dart
@@ -38,7 +38,7 @@ class IPField extends StatelessWidget {
     return SizedBox(
         width: ipWidth,
         child: SpecialTextField(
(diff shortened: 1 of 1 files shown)


nbrownus (approved):
Can do in another or but we should pull out the special bits in IPField that aren't being useful anymore

PR #120, opened 2023-05-10
Update nebula binding code for new APIs

JohnMaguire:
Also ran `go mod tidy`


nebula/mobileNebula.go
```diff
@@ -272,7 +272,7 @@ func VerifyCertAndKey(rawCert string, pemPrivateKey string) (bool, error) {
 		return false, fmt.Errorf("error while unmarshaling cert: %s", err)
 	}
 
-	if err = nebulaCert.VerifyPrivateKey(rawKey); err != nil {
+	if err = nebulaCert.VerifyPrivateKey(cert.Curve_CURVE25519, rawKey); err != nil {
```
JohnMaguire, line 275: I don't think we need to support P256 in the mobile app? If the nebulaCert encodes the curve somewhere I can pass that instead.

jasikpark (commented):
lgtm?

nbrownus (approved):

PR #121, opened 2023-05-11
[feat] Allow for custom server URL's

c0repwn3r:
This PR adds basic support for using a custom API server. It currently only supports iOS, but it would really be trivial to add this to Android too.
I just added a settings bundle, so users can to into Settings -> Nebula -> (Advanced) API Url to change the URL. It still defaults to https://api.defined.net though.


JohnMaguire:
Hi @c0repwn3r - thanks for the contribution. Unfortunately, there are numerous changes and considerations we'd need to make in order to land this feature. In particular iOS-only support is a no-go, and we generally don't use iOS package settings bundles for config.

The comment about making this configurable in `nebula/api.go` was really speaking to a development use case of testing against other API servers (i.e. a build-time flag.) If we did ever open the mobile app up to third-party servers we'd need to make other considerations, like probably enforcing HTTPS, and making this a per-site configuration - which is extra tricky given the way that enrollment URLs work today.

All of this is to say that we're not currently planning to implement this feature and suggest running a fork if it's something you need or desire. Alternatively, we may be open to a pull request which implements server configuration as a build-time flag. Thanks for your understanding!

c0repwn3r:
Sounds good to me, thanks for your consideration. I'll maintain my fork for others who wish to use this feature.

PR #132, opened 2023-05-23
Rework the ios startup flow slightly to avoid crashing on ios 16.5

### LINKED ISSUE #131: On iOS 16.5, the VPN cannot be activated
I have Mobile Nebula installed on both my iPhone and iPad, but after I upgraded to iOS 16.5, the VPN didn't work anymore. Status has always been Disconnect and I can never activate it.

It's happening on at least two of my devices.
Can you give some advice? Thanks.

nbrownus:
iOS 16.5 changed the IPC behavior with the vpn process which caused the process to crash if it received an `appMessage` before it was "ready".

This change no longer caches the system `completionHandler` when being asked to start by the UI, instead calling it immediately, which appears to get the vpn process in the "ready" state. 

This also changes when the UI side determines its safe to send an `appMessage`, this will delay the ui toggle flip from off -> on by a little bit.

There is an opportunity to make the experience feel snappy like pre 16.5 days but it will require a larger lift.

Closes #131

### DIFF
--- a/ios/NebulaNetworkExtension/PacketTunnelProvider.swift
+++ b/ios/NebulaNetworkExtension/PacketTunnelProvider.swift
@@ -13,15 +13,12 @@ class PacketTunnelProvider: NEPacketTunnelProvider {
     private var didSleep = false
     private var cachedRouteDescription: String?
     
-    // This is the system completionHandler, only set when we expect the UI to ask us to actually start so that errors can flow back to the UI
-    private var startCompleter: ((Error?) -> Void)?
-    
-    override func startTunnel(options: [String : NSObject]?, completionHandler: @escaping (Error?) -> Void) {       
+    override func startTunnel(options: [String : NSObject]?, completionHandler: @escaping (Error?) -> Void) {
         // There is currently no way to get initialization errors back to the UI via completionHandler here
         // `expectStart` is sent only via the UI which means we should wait for the real start command which has another completion handler the UI can intercept
-        // In the end we need to call this completionHandler to inform the system of our state
         if options?["expectStart"] != nil {
-            startCompleter = completionHandler
+            // The system completion handler must be called before IPC will work
+            completionHandler(nil)
             return
         }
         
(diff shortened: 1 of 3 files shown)


nbrownus:
@JohnMaguire had asked how things could be improved in the future out of band. The approach I was imagining was splitting the UI on/off toggle state and the description state a further from the IPC command readiness for both the swift and iOS side.

- UI: vpn `connecting` = ok to flip the toggle
- Swift: vpn `connected` = ok to issue start command
- Swift/UI: start command issued and responded to = ok to send further ipc commands

ios/Runner/Sites.swift
```diff
@@ -134,22 +134,22 @@ class SiteUpdater: NSObject, FlutterStreamHandler {
...
             }
-
+            
             self.update(connected: self.site.connected!)
         }
 #endif
         return nil
     }
-
+
```
JohnMaguire, line 152: nit: changes in this file look unrelated and we could drop them from the PR.

ios/NebulaNetworkExtension/PacketTunnelProvider.swift
```diff
@@ -13,15 +13,12 @@ class PacketTunnelProvider: NEPacketTunnelProvider {
...
     
-    // This is the system completionHandler, only set when we expect the UI to ask us to actually start so that errors can flow back to the UI
-    private var startCompleter: ((Error?) -> Void)?
-    
-    override func startTunnel(options: [String : NSObject]?, completionHandler: @escaping (Error?) -> Void) {       
+    override func startTunnel(options: [String : NSObject]?, completionHandler: @escaping (Error?) -> Void) {
         // There is currently no way to get initialization errors back to the UI via completionHandler here
         // `expectStart` is sent only via the UI which means we should wait for the real start command which has another completion handler the UI can intercept
-        // In the end we need to call this completionHandler to inform the system of our state
         if options?["expectStart"] != nil {
```
JohnMaguire, line 19: Is there any advantage at all to keeping these two code paths right now? Or should we just remove the concept of expectStart altogether for now?

JohnMaguire (approved):
Approving, but would re-approve if either comment is addressed.

PR #151, opened 2024-01-31
Fix Github PAT appearing in Android and iOS app builds

johnmaguire:
Also adds a regression test.


nbrownus (approved):
I have a personal preference for `grep -qF "${TOKEN}"` over `grep -F "${TOKEN}" &>/dev/null` but its working so 👍

PR #160, opened 2024-09-03
Update Flutter, target android SDK 34

IanVS:
This updates flutter to 3.24.1, the latest stable version, and also updates our flutter dependencies to latest.  

It targets the latest android sdk, 34, which is required if we want to publish a new version to the Google Play store.  

I also needed to make a few adjustments to handle deprecations.  The biggest change is that I needed to wrap the main widget in `MaterialApp` to avoid problems with AdaptiveSwitch in iOS.


jasikpark (approved):
I don't know enough to give a proper review, but nothing is obviously strange here 👍

ios/Flutter/AppFrameworkInfo.plist
```diff
@@ -21,6 +21,6 @@
   <key>CFBundleVersion</key>
   <string>1.0</string>
   <key>MinimumOSVersion</key>
-  <string>11.0</string>
+  <string>12.0</string>
```
johnmaguire, line 24: Curious why this needed to bump?

lib/components/SpecialTextField.dart
```diff
@@ -90,7 +90,7 @@ class _SpecialTextFieldState extends State<SpecialTextField> {
         autofocus: widget.autofocus,
         focusNode: widget.focusNode,
         onChanged: widget.onChanged,
-        enabled: widget.enabled,
+        enabled: widget.enabled ?? false, // TODO: maybe there's a better way to do this?  widget.enabled is nullable, should it be?
```
johnmaguire, line 93: probably not - the only reason i can think of for it being nullable is if you wanted enabled to be the default state when not specified (but this code defaults to false when unspecified)

lib/services/settings.dart
```diff
@@ -37,14 +37,6 @@ class Settings {
     _set('logWrap', enabled);
   }
 
-  String _getString(String key, String defaultValue) {
```
johnmaguire, line 40: Any reason to remove this?

johnmaguire (commented):
Consider doing `dart fmt` separately and adding the hash to `.git-blame-ignore-revs`. We haven't done this for all of them, but they can be pretty major, as in this PR...

Looks like `go get` ran, so maybe run `go mod tidy` as well to clean up superfluous items in `go.sum`.

lib/services/settings.dart
IanVS: It wasn't being used, and gave a linter error/warning.

johnmaguire: I think even though it's not being used currently, it's worth keeping around, even if a dead-code linter is firing. I don't think it's far from the realm of possibilities that we'll want to store a string value from the settings in the future.

ios/Flutter/AppFrameworkInfo.plist
IanVS: Looks like it's the minimum for xcode 15.4, which is what I'm using.

image

https://developer.apple.com/support/xcode/

johnmaguire: Did some Googling and iOS 11 looks dead as a doorknob. I think this is fine.

johnmaguire:
Mind running `go mod tidy`?

IanVS:
>Mind running go mod tidy?

I assumed that's what happened when I ran `flutter clean` and the `go.mod` file was changed. I wonder if there's a way to extend it to avoid requiring a manual step...

johnmaguire (approved):

IanVS:
Thanks for the review John.  Let's hold off on a release / beta build for now, there's a bit more that I nearly have ready to push up.

PR #164, opened 2024-09-24
Support android API 34

IanVS:
These parameters are now required in android api level 34.  See https://developer.android.com/about/versions/14/behavior-changes-14#runtime-receivers-exported.

I tested by enabling a "good site" in the emulator, but it would be great if someone could test that vpn functionality still does work as expected on a real phone.

### DIFF
--- a/android/app/src/main/kotlin/net/defined/mobile_nebula/NebulaVpnService.kt
+++ b/android/app/src/main/kotlin/net/defined/mobile_nebula/NebulaVpnService.kt
@@ -213,7 +213,7 @@ class NebulaVpnService : VpnService() {
     }
 
     private fun registerReloadReceiver() {
-        registerReceiver(reloadReceiver, IntentFilter(ACTION_RELOAD))
+        registerReceiver(reloadReceiver, IntentFilter(ACTION_RELOAD), RECEIVER_NOT_EXPORTED)
     }
 
     private fun unregisterReloadReceiver() {
--- a/android/app/src/main/kotlin/net/defined/mobile_nebula/MainActivity.kt
+++ b/android/app/src/main/kotlin/net/defined/mobile_nebula/MainActivity.kt
@@ -98,7 +98,7 @@ class MainActivity: FlutterActivity() {
 
         apiClient = APIClient(context)
 
-        registerReceiver(refreshReceiver, IntentFilter(ACTION_REFRESH_SITES))
+        registerReceiver(refreshReceiver, IntentFilter(ACTION_REFRESH_SITES), RECEIVER_NOT_EXPORTED)
 
         enqueueDNUpdater()
     }


johnmaguire:
Huh. Surprised you didn't need to do this in #160. Did the app build then? Does it build now?

johnmaguire (approved):

IanVS:
It built for emulated phones running API level 33, but not 34.  It builds for 34 now, so long as they have a 4kb page size.

android/app/src/main/kotlin/net/defined/mobile_nebula/NebulaVpnService.kt
```diff
@@ -209,11 +209,11 @@ class NebulaVpnService : VpnService() {
             }
         }
 
-        registerReceiver(receiver, IntentFilter(PowerManager.ACTION_DEVICE_IDLE_MODE_CHANGED))
```
johnmaguire, line 212: Is this one actually necessary?

> Exception for receivers that receive only system broadcasts

> If your app is registering a receiver only for [system broadcasts](https://developer.android.com/guide/components/broadcasts#system-broadcasts) through Context#registerReceiver methods, such as [Context#registerReceiver()](https://developer.android.com/reference/android/content/Context#registerReceiver(android.content.BroadcastReceiver,%20android.content.IntentFilter)), then it shouldn't specify a flag when registering the receiver.

johnmaguire: https://android.googlesource.com/platform/prebuilts/fullsdk/platforms/+/9698c0ff01d4b3e24cab840903c06a282db785b3/android-34/data/broadcast_actions.txt?pli=1#

`android.os.action.DEVICE_IDLE_MODE_CHANGED`

I think not?

johnmaguire:
> It built for emulated phones running API level 33, but not 34. It builds for 34 now, so long as they have a 4kb page size.

I hate this stuff. I don't understand how we can targetSdkVersion 34 but still have a build succeed when the app doesn't satisfy the requirements for SDK level 34.

I ask because we always want to keep `master` buildable.

android/app/src/main/kotlin/net/defined/mobile_nebula/NebulaVpnService.kt
jasikpark: It sounds like the app doesn't built on API 34 without this flag and does with the addition of the flag though, right?

johnmaguire: @jasikpark That's not my reading of the first text I quoted, no.

jasikpark: ~~Sorry what I'm saying is it does sound like we don't need the flag from what the docs claim, but since all this does is change those flags, it isn't needed.~~

Realized that this was the unique receiver from the other two. disregard 🫡

IanVS: I'll test it out tomorrow.

jasikpark (approved):

android/app/src/main/kotlin/net/defined/mobile_nebula/NebulaVpnService.kt
IanVS: you were right, seems to work fine without it.

johnmaguire (approved):

johnmaguire:
> Huh. Surprised you didn't need to do this in https://github.com/DefinedNet/mobile_nebula/pull/160. Did the app build then? Does it build now?

For posterity, these generate RuntimeExceptions, not build-time errors. Very annoying.

PR #169, opened 2024-09-27
Add smoke-build CI workflow

IanVS:
This adds a workflow into PRs to build the iOS and Android apps, just to verify that they aren't completely broken / unable to build.  They're debug builds, so it won't completely mirror the release build, but that allows us to avoid getting signing keys in the workflow.


jasikpark:
bravo 🎉

jasikpark (approved):
lgtm, without knowledge of how this all works, but it seems reasonable, plus green CI looks good 👍

ios/fastlane/Fastfile
```diff
@@ -18,6 +18,50 @@ default_platform(:ios)
 platform :ios do
   desc "Push a new beta build to TestFlight"
 
+  lane :checkBuild do
```
nbrownus, line 21: Do we need the signing credentials available? If yes, if we switch this to debug can we avoid it?

nebula/Makefile
```diff
@@ -9,7 +9,7 @@ clean:
 
 mobileNebula.aar: *.go go.sum
 	go get -d golang.org/x/mobile/cmd/gomobile
-	gomobile bind -trimpath -v --target=android -androidapi 26
+	gomobile bind -trimpath -v --target=android -androidapi=26
```
IanVS, line 12: this is dumb but might as well keep it consistent with `--target`

IanVS:
I wish we had this prior to the last PR being merged, which upgraded nebula but caused the build to break.  Working on fixing it now, but would love a review on this PR in the meantime.

nbrownus (approved):

PR #170, opened 2024-10-01
Remove media and audio pickers

IanVS:
And target iphoneos 12, to avoid some warnings. 

We only use the file picker to allow users to load in nebula certs/keys, we don't need access to pictures.


IanVS:
https://github.com/miguelpruivo/flutter_file_picker/issues/1236

ios/Runner/Info.plist
```diff
@@ -43,8 +43,6 @@
 	<true/>
 	<key>NSCameraUsageDescription</key>
 	<string>Camera permission is required for qr code scanning.</string>
-	<key>NSPhotoLibraryUsageDescription</key>
```
jasikpark, line 46: is it possible to load a cert from a picture you took earlier of the qr code?

nbrownus (approved):

ios/Runner/Info.plist
nbrownus: When this was included we didn't have an option to ignore the photo library

jasikpark: so i'm guessing the answer is no then?

nbrownus: I'm sure it could be technically possible but its not a flow I am particularly interested in preserving

PR #171, opened 2024-10-07
Update proguard rules for missing classes

IanVS:
Addresses the errors seen in https://github.com/DefinedNet/mobile_nebula/actions/runs/11221064890/job/31190581077#step:13:1260

Locally I am able to perform most of a release build after removing these proguard rules which were previously added to avoid minification stripping out flutter code.  That seems to no longer be required, though.


johnmaguire:
Can you explain the solution / why these warnings are safe to ignore?

Here it says this could lead to runtime errors if these are actually necessary: https://stackoverflow.com/a/78886649/242129

There's a different solution here that I don't have time to dig into: https://stackoverflow.com/a/76806805/242129

Thanks!

johnmaguire:
I have only glanced at this, but from the SO, the existing proguard-rules.pro entries may be related to the issue? It looks like proguard-rules.pro comes from guidance from https://medium.com/@swav.kulinski/flutter-and-android-obfuscation-8768ac544421 ...

But in https://github.com/flutter/flutter/issues/78625#issuecomment-809915540 it sounds like a bug was fixed - the workaround to that bug was to add these lines to proguard-rules.pro. So maybe the entire proguard-rules.pro is no longer necessary?

Of course, I am probably completely misunderstanding.

IanVS:
I'll experiment with it, I thought this would be safe to ignore because AFAIK we don't use google play apis for anything.  But it looks like https://pub.dev/packages/flutter_plugin_android_lifecycle might take care of minification/obfuscation issues now.

IanVS:
@johnmaguire I think you're right, I was able to clean up the rules and avoid the errors I was seeing previously.

johnmaguire (approved):

PR #181, opened 2024-10-16
Update nebula control for netip

IanVS:
The update to nebula 1.9.4 brought in https://github.com/slackhq/nebula/pull/1173, which changes the way some of the internals of nebula work, which means we need to make some adjustments here as well.  

WIP.  I've gotten gomobile building correctly now, but some kotlin changes are still necessary for interfacing with it.


johnmaguire:
Superseded by #182.

PR #186, opened 2024-10-17
Handle managed nebula updates correctly

IanVS:
This fixes a few issues:

1) When updates are made, we will no longer create duplicate VPN profiles, rather we will update existing ones.
2) We will correctly update an existing profile when the site is running and an update is received, rather than attempting to create a new profile, which failed due to permissions errors.  
3) We will always reload nebula, even if we can't successfully save the VPN profile.
4) The default polling interval of 15 minutes is restored (previously set to 30 seconds during testing).

So far in manual testing I've confirmed that I do not lose the tunnel to my lighthouse even after the original 30 minute expiration of a certificate.  This confirms that reloads are occurring correctly.  Additionally, duplicate sites are not created when updates occur while the site is disconnected.

### DIFF
--- a/ios/NebulaNetworkExtension/PacketTunnelProvider.swift
+++ b/ios/NebulaNetworkExtension/PacketTunnelProvider.swift
@@ -28,76 +28,96 @@ class PacketTunnelProvider: NEPacketTunnelProvider {
     }
     
     private func start(completionHandler: @escaping (Error?) -> Void) {
-        let proto = self.protocolConfiguration as! NETunnelProviderProtocol
-        var config: Data
-        var key: String
-
-        do {
-            site = try Site(proto: proto)
-            config = try site!.getConfig()
-        } catch {
-            //TODO: need a way to notify the app
-            log.error("Failed to render config from vpn object")
-            return completionHandler(error)
-        }
-
-        let _site = site!
-
-        do {
-            key = try _site.getKey()
-        } catch {
-            return completionHandler(error)
-        }
-        
-        let fileDescriptor = tunnelFileDescriptor
-        if fileDescriptor == nil {
-            return completionHandler("Unable to locate the tun file descriptor")
-        }
-        let tunFD = Int(fileDescriptor!)
-
-        // This is set to 127.0.0.1 because it has to be something..
-        let tunnelNetworkSettings = NEPacketTunnelNetworkSettings(tunnelRemoteAddress: "127.0.0.1")
-
-        // Make sure our ip is routed to the tun device
-        var err: NSError?
-        let ipNet = MobileNebulaParseCIDR(_site.cert!.cert.details.ips[0], &err)
-        if (err != nil) {
-            return completionHandler(err!)
-        }
-        tunnelNetworkSettings.ipv4Settings = NEIPv4Settings(addresses: [ipNet!.ip], subnetMasks: [ipNet!.maskCIDR])
-        var routes: [NEIPv4Route] = [NEIPv4Route(destinationAddress: ipNet!.network, subnetMask: ipNet!.maskCIDR)]
-
-        // Add our unsafe routes
-        _site.unsafeRoutes.forEach { unsafeRoute in
-            let ipNet = MobileNebulaParseCIDR(unsafeRoute.route, &err)
-            if (err != nil) {
-                return completionHandler(err!)
-            }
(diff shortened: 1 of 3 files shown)


ios/NebulaNetworkExtension/PacketTunnelProvider.swift
```diff
@@ -28,76 +28,96 @@ class PacketTunnelProvider: NEPacketTunnelProvider {
...
-        do {
-            site = try Site(proto: proto)
-            config = try site!.getConfig()
-        } catch {
-            //TODO: need a way to notify the app
-            log.error("Failed to render config from vpn object")
-            return completionHandler(error)
-        }
+        // Load vpn configs from system, and find the manager matching the one being started
+        NETunnelProviderManager.loadAllFromPreferences() { managers, error in
```
nbrownus, line 32: It would be a bit more readable to chain this in before `start()`

ios/NebulaNetworkExtension/PacketTunnelProvider.swift
```diff
@@ -28,76 +28,96 @@ class PacketTunnelProvider: NEPacketTunnelProvider {
...
+            let targetProtoConfig = self.protocolConfiguration as? NETunnelProviderProtocol;
+            let targetID = targetProtoConfig?.providerConfiguration!["id"] as? String;
```
nbrownus, line 34: It would be more ideal if we didn't explode when trying to get the `id` here and in the loop below and instead called `completionHandler(with the error)`

ios/NebulaNetworkExtension/PacketTunnelProvider.swift
```diff
@@ -28,76 +28,96 @@ class PacketTunnelProvider: NEPacketTunnelProvider {
...
+            
+            guard let managers = managers else {
+                self.log.error("No managers were loaded")
```
nbrownus, line 37: We should call `completionHandler(with an error)` here

ios/NebulaNetworkExtension/PacketTunnelProvider.swift
```diff
@@ -28,76 +28,96 @@ class PacketTunnelProvider: NEPacketTunnelProvider {
...
+            }
+            
+            for manager in managers {
+                var config: Data
+                var key: String
+                
+                let mgr = manager.protocolConfiguration as? NETunnelProviderProtocol;
+                let id = mgr?.providerConfiguration!["id"] as? String;
+                
+                if (id == targetID) {
```
nbrownus, line 48: If we find our `id` then we can stop looping and finish starting, we can close this loop a lot sooner and avoid the `break` at the bottom

ios/NebulaNetworkExtension/PacketTunnelProvider.swift
```diff
@@ -28,76 +28,96 @@ class PacketTunnelProvider: NEPacketTunnelProvider {
...
+                        }
+                        
+                        self.nebula!.start()
+                        self.dnUpdater.updateSingleLoop(site: self.site!, onUpdate: self.handleDNUpdate)
+                        
+                        completionHandler(nil)
+                    })
+                    
+                    // We're done looking through managers
+                    continue
```
nbrownus, line 117: break

IanVS:
@nbrownus thanks for the review.  I tried to refactor this for better readability, using structured concurrency rather than the older completionHandler syntax.  Let me know what you think.

ios/NebulaNetworkExtension/PacketTunnelProvider.swift
```diff
@@ -116,6 +114,25 @@ class PacketTunnelProvider: NEPacketTunnelProvider {
 //        completionHandler()
 //    }
     
+    private func findManager() async throws -> NETunnelProviderManager {
+        let targetProtoConfig = self.protocolConfiguration as? NETunnelProviderProtocol;
+        let targetID = targetProtoConfig?.providerConfiguration!["id"] as? String;
```
nbrownus, line 119: Would be nice if we could avoid blowing up on `targetProtoConfig?.providerConfiguration!` but should be pretty obvious if we stop using our provided config with an `id` in favor of something else.

nbrownus (dismissed):

ios/NebulaNetworkExtension/PacketTunnelProvider.swift
IanVS: I added some guards and a note in `site.swift` where we set the providerConfig and id.

nbrownus (approved):

PR #189, opened 2024-10-23
Add Sentry for error tracking

IanVS:


johnmaguire (approved):

johnmaguire:
Approved - one thing that gives me a small amount of anxiety is that anyone checking out this repo and making their own build (possibly with local changes) will have Sentry error reporting configured? It'd be nice to keep the "signal-to-noise" ratio high. One way to achieve this might be to inject the Sentry DSN in CI during the release build... IDK.

lib/main.dart
```diff
@@ -11,14 +11,26 @@ import 'package:mobile_nebula/screens/MainScreen.dart';
...
 import 'package:flutter_web_plugins/url_strategy.dart';
+import 'package:sentry_flutter/sentry_flutter.dart';
 
-//TODO: EventChannel might be better than the stream controller we are using now
-
-void main() {
+Future<void> main() async {
   usePathUrlStrategy();
-  runApp(Main());
+  await SentryFlutter.init(
```
jasikpark, line 18: Is it possible to add sourcemaps here? Does Dart have sourcemaps / debug symbols? It'd be great to improve the stacktraces in Sentry if possible

jasikpark (approved):
lgtm, the test error showed up fine

lib/main.dart
IanVS: According to https://docs.sentry.io/platforms/flutter/upload-debug/#when-to-upload, there's no "flutter sourcemaps" that we need to upload, but we may want to consider uploading debug symobols for the native android/ios code.  I think that can be done in a followup.

IanVS:
@johnmaguire it's a good point about the DSN.  I started adding it via env var, which can be done, but it does add a layer of complexity.  I think I'd rather run with this as-is for now, and then if we start to see reports from a fork, (which we should be able to determine based on release number, if nothing else), we can rotate our DSN and start using it via env var at that point.

Issue #150, opened 2024-01-24
Add apk to releases


WillPower3309:
Would it be possible to start including a .APK file in new releases of the app? That would provide Android users an easy way to install from release page, and could be very useful for users that like to download their apps from releases through Obtainium (https://github.com/ImranR98/Obtainium/)

WillPower3309:
This would also allow users who do not like having the play store installed (as mentioned by a user in https://github.com/DefinedNet/mobile_nebula/issues/44) to use the app

IanVS:
Included in https://github.com/DefinedNet/mobile_nebula/releases/tag/v0.4.2

PR #190, opened 2024-10-25
Create and upload universal APK during release

IanVS:
This will give users a way to sideload the app without using Google Play or having to deal with running `bundletool`.

Closes https://github.com/DefinedNet/mobile_nebula/issues/150.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -32,6 +32,9 @@ jobs:
         with:
           flutter-version: '3.24.1'
 
+      - name: Setup bundletool for APK generation
+        uses: amyu/setup-bundletool@f7a6fdd8e04bb23d2fdf3c2f60c9257a6298a40a
+
       - name: Install the appstore connect key material
         env:
           AC_API_KEY_SECRET_BASE64: ${{ secrets.AC_API_KEY_SECRET_BASE64 }}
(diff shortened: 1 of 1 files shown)


johnmaguire (approved):
awesome!!

.github/workflows/release.yml
```diff
@@ -32,6 +32,9 @@ jobs:
         with:
           flutter-version: '3.24.1'
 
+      - name: Setup bundletool for APK generation
+        uses: amyu/setup-bundletool@f7a6fdd8e04bb23d2fdf3c2f60c9257a6298a40a
```
johnmaguire, line 36: I don't want to nitpick here, but https://github.com/mukeshsolanki/bundletool-action looks a little more active.

johnmaguire: Some differences - mukeshsolanki downloads the latest bundletool or a specified tag, amyu always downloads 1.15.1.

mukeshsolanki might have a CI warning about using an old node version.

mukeshsolanki is configured a bit differently.

I would need to go read about how CI Actions work, but mukeshsolanki seems to be able to run bundletool for you:

https://github.com/mukeshsolanki/bundletool-action/blob/main/src/action.js#L37-L43

Versus:

https://github.com/amyu/setup-bundletool/blob/main/src/main.ts#L28-L34

Anyway, I don't feel too strongly. I went to go look because using a SHA instead of a version here wasn't typical, and saw that there are no stars or forks. mukeshsolanki is listed first on the Action store with a few stars: https://github.com/marketplace/actions/bundletool-action

So I think it's a wash! But maybe we can stick a comment about why we're using the SHA.

IanVS: I did look at that one briefly, but it requires specifying signing credentials, which made me a bit nervous, and does more than just installing bundletool so I have a feeling it might be less flexible.  This action is really simple and gives us the flexibility to use bundletool however we want, which is why I went with it.

IanVS: For actions that are at all sketchy (not very used, uncommon, etc), I like to use a SHA just to avoid them from releasing something nefarious and us automatically using it.  It just gives us more control over the code that is being run.  I looked over the source code of that commit, so I feel alright using it as-is.

PR #193, opened 2024-10-29
Fix ID used when listening to updates on new sites

IanVS:
This should avoid the errors we've been seeing in Sentry of:

```
MissingPluginException(No implementation found for method listen on channel net.defined.nebula/null)
```

To test, start up the mobile app in an android phone or emulator, tap the plus button to add a new site, and verify that no error log is shown.

Note: we still get a similar error when tapping the "Good Site" or "Bad Site" in debug, but I believe that is because it's being created too quickly before the EventChannel has a chance to be set up.  Adding a site manually does not trigger the error.

### DIFF
--- a/android/app/src/main/kotlin/net/defined/mobile_nebula/MainActivity.kt
+++ b/android/app/src/main/kotlin/net/defined/mobile_nebula/MainActivity.kt
@@ -54,6 +54,7 @@ class MainActivity: FlutterActivity() {
     }
 
     override fun configureFlutterEngine(flutterEngine: FlutterEngine) {
+        super.configureFlutterEngine(flutterEngine)
         appContext = context
         //TODO: Initializing in the constructor leads to a context lacking info we need, figure out the right way to do this
         sites = Sites(flutterEngine)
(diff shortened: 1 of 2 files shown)


lib/models/Site.dart
```diff
@@ -93,7 +93,7 @@ class Site {
     this.rawConfig = rawConfig;
     this.lastManagedUpdate = lastManagedUpdate;
 
-    _updates = EventChannel('net.defined.nebula/$id');
+    _updates = EventChannel('net.defined.nebula/$this.id');
```
johnmaguire, line 96:
```suggestion
    _updates = EventChannel('net.defined.nebula/${this.id}');
```

I could be wrong, but I thought this needed to be curly-braced because of the `.`?

IanVS: 🤦

android/app/src/main/kotlin/net/defined/mobile_nebula/MainActivity.kt
```diff
@@ -54,6 +54,7 @@ class MainActivity: FlutterActivity() {
     }
 
     override fun configureFlutterEngine(flutterEngine: FlutterEngine) {
+        super.configureFlutterEngine(flutterEngine)
```
johnmaguire, line 57: With this change, the call to `GeneratedPluginRegistrant.registerWith(flutterEngine)` can be removed:

https://github.com/flutter/engine/blob/d0ad7cf69bdf78ddda0956b89d2a7d0c9b26fd2c/shell/platform/android/io/flutter/embedding/android/FlutterActivity.java#L1328-L1337

Another ref: https://docs.flutter.dev/platform-integration/platform-channels#step-3-add-an-android-platform-specific-implementation

johnmaguire (approved):

PR #197, opened 2024-11-01
Remove sentry release action

IanVS:
It fails on macOS, and we don't need it.  Sentry automatically constructs a version the first time it sees an error, and the version number it picks seems fine.  

If we decide we do need this in the future (e.g. to associate commits with releases), we'll need to run the sentry cli manually.


jasikpark:
So this doesn't currently associate commits, and it fails on macOS. womp.

jasikpark (approved):

PR #201, opened 2024-12-05
Do not save site if there are errors

### LINKED ISSUE #191: Clicking save on an unvalidated form closes the form without saving
Reported on Nebula OSS Slack: https://nebulaoss.slack.com/archives/CRWJJM52B/p1728932727151009

https://github.com/user-attachments/assets/b26ad0ce-cca3-418d-8816-b84dfb6a0a48

IanVS:
Fixes #191 

This makes Android behave more like iOS, wherein it is not possible to save an invalid site.  As a bonus here, we'll display a more useful error message as well, referencing why the form cannot be saved.

### DIFF
--- a/android/app/src/main/kotlin/net/defined/mobile_nebula/MainActivity.kt
+++ b/android/app/src/main/kotlin/net/defined/mobile_nebula/MainActivity.kt
@@ -189,7 +189,10 @@ class MainActivity: FlutterActivity() {
             return result.error("unhandled_error", err.message, null)
         }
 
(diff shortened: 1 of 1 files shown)


nbrownus (dismissed):
For both iOS and Android it would be ideal to be able to not lose work if you can't complete a fully valid configuration in one run.

IanVS:
I'm going to back out some of this then, and work on showing a warning prompt that can be accepted to save an invalid config.

johnmaguire (approved):

PR #210, opened 2024-12-18
Fix ConfigItem font size

IanVS:
And align config items on the right.

|Before:|After:|Android after:|
|----|----|----|
|image|image|image|

(Android has always had some mismatched text sizes there...)


johnmaguire (approved):

johnmaguire:
Do you have screenshots of the Status page with the "Connected" toggle too? Would be nice to capture for posterity if nothing else.

IanVS:
image

Issue #200, opened 2024-11-13
Investigate Sentry crash reporting opt-out


johnmaguire:
We added Sentry crash reporting in order to try to track down some of the more obscure issues reported by users that do not show up in the Play Developer Console crash reports (e.g. #136, #173, etc.) Some users have privacy concerns with respect to reporting crashes to third-party services (#199) so let's investigate adding an opt-out to the settings.

PR #211, opened 2024-12-20
Add Sentry crash-reporting opt-out

IanVS:
Closes https://github.com/DefinedNet/mobile_nebula/issues/200

|Before|After|
|---|---|
|||


johnmaguire (approved):

nbrownus:
Do we need to tear anything down when disabling?

IanVS:
It won't re-enable if it's toggled on, that will only take effect when the app restarts.  But it disables immediately when toggled off.

PR #212, opened 2024-12-20
Fix system brightness setting

IanVS:
This fixes two issues:

1) Assume the system is set to dark mode.  Previously, if you disabled system color in the app settings, then changed the dark mode switch to off, then turned system color back on, the app would stay light.  Now, it will respect the system color correctly.

2) Previously the app would not change color if the system mode was changed.  Now it listens for system brightness changes, and updates accordingly, if the setting for using system color is turned on.

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -64,6 +64,16 @@ class _AppState extends State<App> {
       });
     });
 
+    // Listen to changes to the system brightness mode, update accordingly
+    var dispatcher = SchedulerBinding.instance.platformDispatcher;
+    dispatcher.onPlatformBrightnessChanged = () {
+      if (settings.useSystemColors) {
+        setState(() {
+          brightness = dispatcher.platformBrightness;
+        });
+      }
+    };
+
     super.initState();
   }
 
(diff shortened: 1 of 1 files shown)


nbrownus (dismissed):

lib/main.dart
```diff
@@ -56,12 +56,24 @@ class _AppState extends State<App> {
...
+        if (settings.useSystemColors) {
+          brightness = SchedulerBinding.instance.platformDispatcher.platformBrightness;
+        } else {
           brightness = settings.darkMode ? Brightness.dark : Brightness.light;
         }
       });
     });
 
+    // Listen to changes to the system brightness mode, update accordingly
+    var dispatcher = SchedulerBinding.instance.platformDispatcher;
```
johnmaguire, line 68: Can this be `val`? I know you are modifying the instance member, but the memory address of the object stored at `dispatcher` doesn't change. Genuine question, I'm not sure how Dart behaves here.

johnmaguire: Uh, I must be thinking of Kotlin. Dart doesn't seem to have val?

johnmaguire (approved):

lib/main.dart
johnmaguire: Oh it'd be `final`. But meh.

johnmaguire (approved):

PR #214, opened 2025-01-09
Update to flutter 3.27

IanVS:
The main thing here is that it will support edge-to-edge when we target android 15, and it changes the default for deep links.

The update to mobile_scanner should improve ios simulator support, but I'm finding that the app still won't load in the simulator (even without mobile_scanner at all), so there's some other work to do there.

https://medium.com/flutter/whats-new-in-flutter-3-27-28341129570c

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -42,7 +42,6 @@
                 <category android:name="android.intent.category.LAUNCHER"/>
             </intent-filter>
             <!-- App linking -->
-            <meta-data android:name="flutter_deeplinking_enabled" android:value="true" />
             <intent-filter android:autoVerify="true">
                 <action android:name="android.intent.action.VIEW"/>
                 <category android:name="android.intent.category.DEFAULT"/>
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ Install all of the following things:
 
 - [`xcode`](https://apps.apple.com/us/app/xcode/)
 - [`android-studio`](https://developer.android.com/studio)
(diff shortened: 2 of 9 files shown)


android/app/src/main/AndroidManifest.xml
```diff
@@ -42,7 +42,6 @@
                 <category android:name="android.intent.category.LAUNCHER"/>
             </intent-filter>
             <!-- App linking -->
-            <meta-data android:name="flutter_deeplinking_enabled" android:value="true" />
```
johnmaguire, line 45: Why are we removing this? We intentionally opted out of Flutter's default deep link handler: https://docs.flutter.dev/ui/navigation/deep-linking#migrating-from-plugin-based-deep-linking

IanVS: >We intentionally opted out of Flutter's default deep link handler

But this was opting it _in_, right?  And that's now the default, so there's no need to keep the flag.  We're not using a plugin-based deep link.  There's some description of the change towards the end of https://medium.com/flutter/whats-new-in-flutter-3-27-28341129570c.

johnmaguire: Derp, thanks. I wonder if the new plugin-based approach would've let us get the raw URL value, and let us solve our upstream issue we ended up working around with a different enrollment code scheme... https://github.com/flutter/flutter/issues/114630

johnmaguire (approved):

johnmaguire:
Ran an iOS build on my machine, and the Simulator hung at a white screen for about 30 seconds then opened. This is the behavior I'm used to in the Simulator.

PR #215, opened 2025-01-10
Allow saving invalid sites

IanVS:
Closes https://github.com/DefinedNet/mobile_nebula/issues/191

@nbrownus would like us to be able to save an invalid site, as a way of saving partial progress and then fixing issues later on, so that's what this PR does, for both Android and iOS.  To test, add a site, give it a name, and press save.  You should be taken to a list of sites, including the newly created invalid site.


johnmaguire:
Went through Git history to see if there was a reason for `getKey` but it has existed since initial commit, so I expect it's an artifact of keeping state clean during development. It should be possible to still delete the site with an invalid key and recreate it. LGTM!

johnmaguire (approved):

PR #222, opened 2025-01-17
Fix iOS 16 support

IanVS:
In older versions of iOS, it's not possible to call `NETunnelProviderManager.loadAllFromPreferences()` from inside the network extension process.  We were seeing `NETunnelProviderManager objects cannot be instantiated from NEProvider processes` errors in iOS 16.  It's unclear exactly when the change happened to allow it, but as far as we can tell it was in iOS 17. 

To Test:
1. On a real device running iOS 16, ensure that enrolling as a Managed Nebula host works correctly.
2. Start the site.
3. Update the host in the admin panel and wait at least 15 minutes for a `checkForUpdate` from the mobile client.  You should get a `Host renewed` audit log for the host.  
4. Verify that there's a log for "Reloading Nebula" in the mobile host, and that it has an up-to-date config.

### DIFF
--- a/ios/NebulaNetworkExtension/PacketTunnelProvider.swift
+++ b/ios/NebulaNetworkExtension/PacketTunnelProvider.swift
@@ -52,14 +52,19 @@ class PacketTunnelProvider: NEPacketTunnelProvider {
         var config: Data
         var key: String
         
-        manager = try await self.findManager()
-        
-        guard let foundManager = manager else {
-            throw VPNStartError.couldNotFindManager
-        }
-        
         do {
-            self.site = try Site(manager: foundManager)
+            // Cannot use NETunnelProviderManager.loadAllFromPreferences() in earlier versions of iOS
+            // TODO: Remove else once we drop support for iOS 16
+            if ProcessInfo().isOperatingSystemAtLeast(OperatingSystemVersion(majorVersion: 17, minorVersion: 0, patchVersion: 0)) {
+                manager = try await self.findManager()
+                guard let foundManager = manager else {
+                    throw VPNStartError.couldNotFindManager
+                }
+                self.site = try Site(manager: foundManager)
+            } else {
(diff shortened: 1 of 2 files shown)


ios/Runner/DNUpdate.swift
```diff
@@ -63,7 +63,10 @@ class DNUpdater {
                 return
             }
             
-            newSite?.save(manager: site.manager) { error in
+            let siteManager = site.manager
+            let shouldSaveToManager = siteManager != nil || ProcessInfo().isOperatingSystemAtLeast(OperatingSystemVersion(majorVersion: 17, minorVersion: 0, patchVersion: 0))
+            
+            newSite?.save(manager: site.manager, saveToManager: shouldSaveToManager) { error in
```
IanVS, line 69: This can be called from the `NETunnelProviderManager` process, in which case there is no manager, and we can't create one in older versions of iOS.

ios/NebulaNetworkExtension/PacketTunnelProvider.swift
```diff
@@ -52,14 +52,19 @@ class PacketTunnelProvider: NEPacketTunnelProvider {
...
+            // Cannot use NETunnelProviderManager.loadAllFromPreferences() in earlier versions of iOS
+            // TODO: Remove else once we drop support for iOS 16
+            if ProcessInfo().isOperatingSystemAtLeast(OperatingSystemVersion(majorVersion: 17, minorVersion: 0, patchVersion: 0)) {
+                manager = try await self.findManager()
+                guard let foundManager = manager else {
+                    throw VPNStartError.couldNotFindManager
+                }
+                self.site = try Site(manager: foundManager)
+            } else {
+                // The downside here is that a new vpn profile is created each time it starts
```
IanVS, line 65: I am not observing this, though I expected it.  I'll remove the comment.

nbrownus (approved):

PR #231, opened 2025-01-23
Use platform-specific canvas color for SimplePage + platform-specific icons

jasikpark:
This uses the `Theme.of(context)` to set the background for SimplePage in a cross-platform respecting way, instead of hardcoding to a cupertino value.

It also uses an icon for the "more" menu that is cross-platform adjusting as well.

| Before | After |
|--------|--------|
| Screenshot_1737669862 | Screenshot_1737669848 |
| Simulator Screenshot - iPhone 16 Pro - 2025-01-23 at 16 04 24 | Simulator Screenshot - iPhone 16 Pro - 2025-01-23 at 16 04 11 |

## Dark mode **after**
| Android | iOS |
|--------|--------|
| Screenshot_1737670925 | Simulator Screenshot - iPhone 16 Pro - 2025-01-23 at 16 23 43 |

### DIFF
--- a/lib/components/SimplePage.dart
+++ b/lib/components/SimplePage.dart
@@ -1,4 +1,3 @@
-import 'package:flutter/cupertino.dart' as cupertino;
 import 'package:flutter/material.dart';
 import 'package:flutter_platform_widgets/flutter_platform_widgets.dart';
 import 'package:mobile_nebula/services/utils.dart';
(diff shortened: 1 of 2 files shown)


nbrownus (approved):
I don't recall the specifics here, I would check that dark mode still looks sane though.

jasikpark:
> I don't recall the specifics here, I would check that dark mode still looks sane though.

dark mode looks good to me!

PR #233, opened 2025-01-23
Add vertical padding to ConfigPageItem in case children wrap

jasikpark:
| Before | After |
|--------|--------|
| Screenshot_1737675596 | Screenshot_1737675583 | 
| Simulator Screenshot - iPhone 16 Pro - 2025-01-23 at 17 44 11 | Simulator Screenshot - iPhone 16 Pro - 2025-01-23 at 17 44 03 |

### DIFF
--- a/lib/components/config/ConfigPageItem.dart
+++ b/lib/components/config/ConfigPageItem.dart
@@ -45,7 +45,7 @@ class ConfigPageItem extends StatelessWidget {
       onPressed: this.disabled ? null : onPressed,
       color: Utils.configItemBackground(context),
       child: Container(
-          padding: EdgeInsets.only(left: 15, right: 15),
+          padding: EdgeInsets.symmetric(horizontal: 15, vertical: 10),
           constraints: BoxConstraints(minHeight: Utils.minInteractiveSize, minWidth: double.infinity),
           child: Row(
             crossAxisAlignment: crossAxisAlignment,


lib/components/config/ConfigPageItem.dart
```diff
@@ -45,13 +45,17 @@ class ConfigPageItem extends StatelessWidget {
...
+          padding: EdgeInsets.symmetric(horizontal: 15, vertical: 10),
           constraints: BoxConstraints(minHeight: Utils.minInteractiveSize, minWidth: double.infinity),
           child: Row(
             crossAxisAlignment: crossAxisAlignment,
             children: <Widget>[
               label != null ? Container(width: labelWidth, child: label) : Container(),
-              Expanded(child: Container(child: content, padding: EdgeInsets.only(right: 10))),
+              Expanded(
+                  child: Container(
+                child: content,
```
IanVS, line 56: Can you run `dart format lib/ test/ -l 120` to clean up the indentation here?  I'd suggest adding it into your android studio (https://stackoverflow.com/a/77387706/1435658).

IanVS: oh wild, this is what the formatter is doing.  That definitely seems like a bug, lol.

IanVS: If you remove the trailing comma at the end of the arguments for `Expanded`, it collapses onto one line, which arguably looks quite a bit better.  But I won't block on that.

IanVS (dismissed):

IanVS (approved):

Issue #12, opened 2021-02-04
Move `logLocalTZ` and `logWrap` to LogScreen.dart


micha3lbrown:
From a UI/UX perspective instead of having some log settings in `AdvancedScreen.dart` and others in `SettingsScreen.dart` it'd be nice to consolidate these features closer to where the logs are displayed. 

While viewing the logs we should add a hamburger menu item in the top right that shows a form that includes the following options: 
- Display local timezone
- Wrap log output 
- [Optional] Filter logs by SEV

PR #235, opened 2025-01-24
Add text wrap toggle to logs screen

jasikpark:
Speaks to https://github.com/DefinedNet/mobile_nebula/issues/12

I've also included stateful tooltips for the button.

Sorry for the lack of logs in the iOS screenshot, I'm not sure why I'm not able to generate a log file in the simulator.

| No wrap | Wrap |
|--------|--------|
| Simulator Screenshot - iPhone 16 Pro - 2025-01-24 at 15 22 49 | Simulator Screenshot - iPhone 16 Pro - 2025-01-24 at 15 22 51 |
| Screenshot_1737753559 | Screenshot_1737753561 |

### DIFF
--- a/lib/screens/SiteLogsScreen.dart
+++ b/lib/screens/SiteLogsScreen.dart
@@ -45,6 +45,31 @@ class _SiteLogsScreenState extends State<SiteLogsScreen> {
 
     return SimplePage(
       title: title,
+      trailingActions: [
+        Padding(
+          padding: const EdgeInsets.only(right: 8),
+          child: IconButton.filledTonal(
+            // Fixes enormous button on iOS, this *should* be the default, but it seems to be overridden.
+            iconSize: 24,
+            style: Platform.isIOS
+                ? IconButton.styleFrom(
+                    shape: RoundedRectangleBorder(borderRadius: BorderRadius.circular(8)),
+                    // Match background when not selected on iOS
+                    backgroundColor: settings.logWrap ? null : Colors.transparent,
+                  )
+                : null,
+            isSelected: settings.logWrap,
+            tooltip: "Turn ${settings.logWrap ? "off" : "on"} text wrapping",
+            selectedIcon: const Icon(Icons.wrap_text_outlined),
+            icon: const Icon(Icons.wrap_text),
+            onPressed: () => {
+              setState(() {
+                settings.logWrap = !settings.logWrap;
+              })
+            },
+          ),
+        )
+      ],
       scrollable: SimpleScrollable.both,
       scrollController: controller,
       onRefresh: () async {
--- a/lib/screens/SiteLogsScreen.dart
+++ b/lib/screens/SiteLogsScreen.dart
@@ -45,31 +45,7 @@ class _SiteLogsScreenState extends State<SiteLogsScreen> {
 
     return SimplePage(
       title: title,
-      trailingActions: [
-        Padding(
-          padding: const EdgeInsets.only(right: 8),
-          child: IconButton.filledTonal(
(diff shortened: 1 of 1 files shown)


IanVS:
You can't actually start up a vpn service in the simulator.  I'd suggest connecting your iphone to your macbook, turning on developer mode, and then running your dev build there, so you can test out actual vpn connections and collect logs.

jasikpark:
On my iOS 15.0.2 iPhone 7:

| No wrap | Wrap |
|--------|--------|
| Screenshot 2025-01-24 at 2 52 43 PM | Screenshot 2025-01-24 at 2 52 48 PM |

lib/screens/SiteLogsScreen.dart
```diff
@@ -45,6 +45,31 @@ class _SiteLogsScreenState extends State<SiteLogsScreen> {
...
+            // Fixes enormous button on iOS, this *should* be the default, but it seems to be overridden.
+            iconSize: 24,
+            style: Platform.isIOS
+                ? IconButton.styleFrom(
+                    shape: RoundedRectangleBorder(borderRadius: BorderRadius.circular(8)),
+                    // Match background when not selected on iOS
+                    backgroundColor: settings.logWrap ? null : Colors.transparent,
+                  )
+                : null,
+            isSelected: settings.logWrap,
```
jasikpark, line 61: This indicates to the button that it is a toggleable button.

jasikpark:
Padding fixed!

jasikpark:
looks like https://medium.com/flutter/whats-new-in-flutter-3-27-28341129570c has some ios changes that might make this simpler.

jasikpark:
Fiddled a bit more with things, I *think* this works better? It's more and less hacky on iOS in different ways.

| No wrap | Wrap |
|--------|--------|
| Simulator Screenshot - iPhone 16 Pro - 2025-01-24 at 16 57 47 | Simulator Screenshot - iPhone 16 Pro - 2025-01-24 at 16 57 45 |
| Screenshot 2025-01-24 at 4 59 17 PM | Screenshot 2025-01-24 at 4 59 23 PM | 
| 
Screenshot_1737759688
 | 
Screenshot_1737759690
 |

lib/screens/SiteLogsScreen.dart
```diff
@@ -64,6 +65,37 @@ class _SiteLogsScreenState extends State<SiteLogsScreen> {
     );
   }
 
+  Widget _buildTextWrapToggle() {
+    return Platform.isIOS
```
IanVS, line 69: We seem to have a bit of a convention that platform ternaries use `isAndroid` instead of `isIOS`.  I don't think it really matters, but we definitely have more android checks than ios, and maybe it's worth sticking to that convention.

lib/screens/SiteLogsScreen.dart
```diff
@@ -64,6 +65,37 @@ class _SiteLogsScreenState extends State<SiteLogsScreen> {
...
+        ? Tooltip(
```
IanVS, line 70: Cool, good to see some accessibility being added.

IanVS (approved):
The difference between enabled and disabled is very subtle in android, but I think it's fine, and seems to be the best button choice from what I can tell.

Did you consider adding the toggle to the bottom bar, along with the other buttons?  Putting it in the app bar leaves less room for the title, but the buttons at the bottom are command buttons, so this toggle might be out of place down there.

jasikpark:
I think keeping this in the top app bar makes the most sense since it's a toggle setting, rather than a stateless action

PR #236, opened 2025-01-27
Update the bottom app bar for the logs screen

jasikpark:
Updated the bottom app sheet for Android and iOS

| iOS | Android |
|--------|--------|
| Simulator Screenshot - iPhone 16 Pro - 2025-01-27 at 11 12 30 | image |


IanVS:
Looks like the Android screenshot is broken.

IanVS (approved):
Looks good.  For posterity, here's the "before" on android:

Android Studio 2025-01-29 09 10 34

And I particularly like that we don't have massive squashed buttons at the bottom anymore. Previously:

Android Studio 2025-01-29 09 10 40

PR #244, opened 2025-01-30
Adjust vertical padding for config items

### LINKED ISSUE #241: Fix extra padding for the Switch in ConfigItems
Image

The two config items are different heights.

jasikpark:
Relevant to #241 

| Before | After |
|--------|--------|
| Screenshot_1738274530 | Screenshot_1738274518 |
| Simulator Screenshot - iPhone 16 Pro - 2025-01-30 at 16 02 11 | Simulator Screenshot - iPhone 16 Pro - 2025-01-30 at 16 02 00 |

### DIFF
--- a/lib/components/config/ConfigItem.dart
+++ b/lib/components/config/ConfigItem.dart
@@ -29,7 +29,7 @@ class ConfigItem extends StatelessWidget {
 
     return Container(
         color: Utils.configItemBackground(context),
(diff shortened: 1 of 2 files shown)


jasikpark:
> It is still a different size from the row below it. Maybe @bgolat has an idea of what we should do here. We can make everything taller, or perhaps special-case just this particular row with the switches to decrease the padding further.

I agree, though maybe we can start with this? I think this moves in the right direction.

IanVS (approved):

PR #245, opened 2025-01-31
Don't pop an error when there is no logs file

jasikpark:
| Before | After |
|--------|--------|
| Screenshot_1738344556 | Screenshot_1738344475 |
| Simulator Screenshot - iPhone 16 Pro - 2025-01-31 at 11 29 51 | Simulator Screenshot - iPhone 16 Pro - 2025-01-31 at 11 27 58 |

### DIFF
--- a/lib/screens/SiteLogsScreen.dart
+++ b/lib/screens/SiteLogsScreen.dart
@@ -5,6 +5,8 @@ import 'package:flutter/material.dart';
 import 'package:flutter_platform_widgets/flutter_platform_widgets.dart';
 import 'package:mobile_nebula/components/SimplePage.dart';
 import 'package:mobile_nebula/models/Site.dart';
+import 'package:mobile_nebula/services/logs.dart';
+import 'package:mobile_nebula/services/result.dart';
 import 'package:mobile_nebula/services/settings.dart';
 import 'package:mobile_nebula/services/share.dart';
 import 'package:mobile_nebula/services/utils.dart';
@@ -22,14 +24,14 @@ class SiteLogsScreen extends StatefulWidget {
 }
 
 class _SiteLogsScreenState extends State<SiteLogsScreen> {
-  String logs = '';
-  ScrollController controller = ScrollController();
-  RefreshController refreshController = RefreshController(initialRefresh: false);
+  final ScrollController controller = ScrollController();
+  final RefreshController refreshController = RefreshController(initialRefresh: false);
+  final LogsNotifier logsNotifier = LogsNotifier();
 
   var settings = Settings();
   @override
   void initState() {
-    loadLogs();
+    logsNotifier.loadLogs(logFile: widget.site.logFile);
     super.initState();
   }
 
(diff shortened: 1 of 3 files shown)


lib/services/logs.dart
```diff
@@ -0,0 +1,31 @@
+import 'dart:io';
+
+import 'package:flutter/widgets.dart';
+import 'package:mobile_nebula/services/result.dart';
+
+class LogsNotFoundException implements Exception {
+  String error() => 'No logs file found';
```
IanVS, line 7: I'm not sure that the word "file" is important to the user here, since that's an implementation detail.  I'd suggest instead we say, "No logs found", or if we want to give the user a hint as to why there are no logs, something like, "No logs found.  Has the site been started?"

lib/screens/SiteLogsScreen.dart
```diff
@@ -143,27 +156,6 @@ class _SiteLogsScreenState extends State<SiteLogsScreen> {
...
-        logs = v;
-      });
-    } on FileSystemException {
-      Utils.popError(context, 'Error while reading logs', 'No log file was present');
-    } catch (err) {
-      Utils.popError(context, 'Error while reading logs', err.toString());
-    }
-  }
-
-  deleteLogs() async {
```
IanVS, line 161: Was this unused?

lib/screens/SiteLogsScreen.dart
```diff
@@ -49,18 +51,29 @@ class _SiteLogsScreenState extends State<SiteLogsScreen> {
...
           padding: EdgeInsets.all(5),
           constraints: logBoxConstraints(context),
-          child: SelectableText(logs.trim(), style: TextStyle(fontFamily: 'RobotoMono', fontSize: 14))),
+          child: ListenableBuilder(
+            listenable: logsNotifier,
+            builder: (context, child) => SelectableText(
+                switch (logsNotifier.logsResult) {
+                  Ok<String>(:var value) => value.trim(),
+                  Error<String>(:var error) => error is LogsNotFoundException
+                      ? "No logs file found."
```
IanVS, line 71: Why not use the exception's text here, which is the same, but without a period at the end?

lib/screens/SiteLogsScreen.dart
jasikpark: Yep!

lib/screens/SiteLogsScreen.dart
jasikpark: Sounds good

lib/services/logs.dart
```diff
@@ -0,0 +1,31 @@
+import 'dart:io';
+
+import 'package:flutter/widgets.dart';
+import 'package:mobile_nebula/services/result.dart';
+
+class LogsNotFoundException implements Exception {
+  String error() => 'No logs found. Has the site been started?';
```
IanVS, line 7: @nbrownus, you good with this wording?

nbrownus: The question feels weird to me, unless there was some catastrophic error during a site start then there would be at least 1 log line in a file to read. 

"Logs will be available after starting the site for the first time" or something feels more helpful

IanVS (approved):

PR #251, opened 2025-02-12
Upgrade to Flutter 3.29.0

jasikpark:
https://medium.com/flutter/whats-new-in-flutter-3-29-f90c380c2317

Gets us the new Material 3 screen animations and some fidelity improvements w/ how native things look.

- https://main-api.flutter.dev/flutter/material/FadeForwardsPageTransitionsBuilder-class.html
- https://medium.com/dartlang/announcing-dart-3-7-bf864a1b195c

PR #252, opened 2025-02-13
Flutter formatting changes

jasikpark:
Follow on to #251 

This PR will get a `.git-blame-ignore-revs` entry after merging.

### DIFF
--- a/.github/workflows/flutterfmt.yml
+++ b/.github/workflows/flutterfmt.yml
@@ -23,6 +23,5 @@ jobs:
         with:
           show-progress: false
 
-      # Disabled for a single PR, this will be re-enabled in the PR that has lots of formatting changes.
-      # - name: Check formating
-      #   run: dart format -l120 lib/ --set-exit-if-changed --suppress-analytics --output none
+      - name: Check formating
+        run: dart format -l120 lib/ --set-exit-if-changed --suppress-analytics --output none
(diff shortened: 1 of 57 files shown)


pubspec.yaml
```diff
@@ -14,7 +14,7 @@ description: Mobile Nebula Client
 version: 0.1.0+54
 
 environment:
-  sdk: ^3.5.1
+  sdk: ^3.7.0
```
johnmaguire, line 17: If you're planning to gitignore this, it should probably be only formatting changes, not version updates or anything functional.

.github/workflows/flutterfmt.yml
```diff
@@ -9,16 +9,14 @@ on:
...
+  flutterfmt:
     name: Run flutter format
     runs-on: ubuntu-latest
     steps:
-
       - name: Install flutter
         uses: subosito/flutter-action@v2
         with:
-          flutter-version: '3.27.0'
+          flutter-version: '3.29.0'
```
johnmaguire, line 19: Same thought here. Though it seems like this was maybe just missed from the last PR?

pubspec.yaml
jasikpark: rebase on main fixed this diff, github is weird sometimes...

.github/workflows/flutterfmt.yml
jasikpark: rebase on main fixed this diff, github is weird sometimes...

johnmaguire (approved):

PR #256, opened 2025-02-13
Add Swift formatting locally and in CI

jasikpark:
https://github.com/nicklockwood/SwiftFormat automatically formats swift code like how we treat dart and golang.

#254 was a different attempt, but https://github.com/swiftlang/swift-format doesn't support easily ignoring a source folder, and the "is it formatted" command also detects lints.

Refs: 
- https://github.com/swiftlang/swift-format/issues/870
- https://github.com/swiftlang/swift-format/pull/873
- https://github.com/swiftlang/swift-format/issues/931

There's an option to load the tool via Cocoapods, or Swift Package Manager, but I think just allowing users to install via homebrew, or whichever method they wish is best for now.

The actual CI check is disabled, enabled in #257

### DIFF
--- a/.github/workflows/swiftfmt.yml
+++ b/.github/workflows/swiftfmt.yml
@@ -17,5 +17,6 @@ jobs:
         with:
           show-progress: false
 
-      - name: Check formating
-        run: swiftformat --lint . --reporter github-actions-log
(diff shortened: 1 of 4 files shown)


jasikpark:
closing for #261

PR #267, opened 2025-02-25
Use material 3 page transitions [android]

IanVS:
The latest version of flutter (3.29, which we're already using) finally added a decent page transition matching up with the native material 3 transition.  (https://main-api.flutter.dev/flutter/material/FadeForwardsPageTransitionsBuilder-class.html)

To test, fire up the app in android, and navigate around.


jasikpark (approved):
nice, meant to do this when upgrading flutter, but I figured it'd be a default we needed to enable, not a new custom thing

PR #268, opened 2025-03-04
Fix some `flutter analyze` warnings

jasikpark:
There's a bunch of style opinions that I didn't fix, plus a potentially more worrisome warning about passing `context` across async gaps, but there's just a few changes here that improve things slightly.

image

### DIFF
--- a/lib/components/CIDRField.dart
+++ b/lib/components/CIDRField.dart
@@ -2,6 +2,7 @@ import 'package:flutter/cupertino.dart';
 import 'package:flutter/services.dart';
 import 'package:mobile_nebula/components/SpecialTextField.dart';
 import 'package:mobile_nebula/models/CIDR.dart';
+
 import '../services/utils.dart';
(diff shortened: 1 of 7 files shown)


lib/components/CIDRField.dart
```diff
@@ -52,57 +53,55 @@ class _CIDRFieldState extends State<CIDRField> {
...
+              textAlign: TextAlign.end,
+              focusNode: widget.focusNode,
+              nextFocusNode: bitsFocus,
               onChanged: (val) {
                 if (widget.onChanged == null) {
                   return;
                 }
 
-                cidr.bits = int.tryParse(val) ?? 0;
+                cidr.ip = val;
```
jasikpark, line 74: This diff is confusing, all that happened in this file is removing an empty wrapping `Container`, this text matches but is just visible lower in the diff

IanVS (approved):

PR #271, opened 2025-03-24
Update to flutter stable 3.29.2

jasikpark:
https://github.com/flutter/flutter/blob/stable/CHANGELOG.md#3292


IanVS:
@jasikpark are you still working on this?

jasikpark:
it's good to go

IanVS:
Why the change to double quotes?

jasikpark:
> Why the change to double quotes?

VSCode find and replace ran my YAML formatter. Is it okay if we keep the changes?

IanVS:
I think it would be better to avoid unrelated formatting changes that are not enforced by the project, to keep the git blame clean.

jasikpark:
image
looks like setting this in my vscode user settings allowed me to find and replace without the formatting 👍

IanVS (approved):

PR #293, opened 2025-11-19
OIDC

### LINKED ISSUE #310: [feature request] Add SSO endpoint device connection in mobile apps as well
As with the latest release of DNClient Desktop support for SSO enddevice authentication - extending that feature to the nebula iOS and android apps makes sense.

JackDoan:
Closes #310

### DIFF
--- a/android/app/src/main/kotlin/net/defined/mobile_nebula/NebulaVpnService.kt
+++ b/android/app/src/main/kotlin/net/defined/mobile_nebula/NebulaVpnService.kt
@@ -339,9 +339,9 @@ class NebulaVpnService : VpnService() {
         }
 
         private fun listIndexes(m: NebulaVpnService, msg: Message) {
-            if (protect(msg)) { return }
+            if (protect(m, msg)) { return }
 
-            val res = nebula!!.listIndexes(false)
+            val res = m.nebula!!.listIndexes(false)
             val m = Message.obtain(null, msg.what)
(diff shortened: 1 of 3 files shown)


pubspec.yaml
```diff
@@ -36,7 +36,7 @@ dependencies:
   flutter_svg: ^2.0.10+1
   intl: ^0.19.0
   share_plus: ^10.0.2
-  sentry_flutter: ^8.14.2
+  sentry_flutter: ^9.8.0
```
JackDoan: this breaks the current iOS CI build, but makes Android shut up about deprecations :disappointed:

android/app/src/main/kotlin/net/defined/mobile_nebula/MainActivity.kt
```diff
@@ -75,6 +80,9 @@ class MainActivity: FlutterActivity() {
                 "nebula.verifyCertAndKey" -> nebulaVerifyCertAndKey(call, result)
 
                 "dn.enroll" -> dnEnroll(call, result)
+                "dn.getPollToken" -> dnGetPollToken(call, result)
```
JackDoan: if you want to play with this on iOS, these are the bindings you need to implement.

Below, you can see I also had to make a version of the binding channel that runs on the background executor, to avoid blocking the UI thread. You probably need both.

PR #296, opened 2025-11-26
Fix crashes in go with composite action

JackDoan:


JackDoan:
Fix confirmed on Android!

PR #299, opened 2025-12-03
switch to LongPollWait

JackDoan:


nebula/api.go
```diff
@@ -104,18 +104,27 @@ func (c *APIClient) TryUpdate(siteName string, hostID string, privateKey string,
...
 	switch {
 	case errors.As(err, &dnapi.ErrInvalidCredentials):
 		return nil, InvalidCredentialsError{}
 	case err != nil:
 		return nil, fmt.Errorf("CheckForUpdate error: %s", err)
 	}
-
-	if !updateAvailable {
+	var msgType struct{ Command string }
+	err = json.Unmarshal(msg.Action, &msgType)
```
johnmaguire: Dang we return a raw JSON from dnapi for this? That's kinda ugly.

JackDoan: strong types across the language boundary is kinda poorly supported, and without the alignment patch, would crash sometimes :cry:

nebula/api.go
```diff
@@ -104,18 +104,27 @@ func (c *APIClient) TryUpdate(siteName string, hostID string, privateKey string,
 	// Check for update
 	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
 	defer cancel()
-	updateAvailable, err := c.c.CheckForUpdate(ctx, creds)
+	msg, err := c.c.LongPollWait(ctx, creds, []string{"DoUpdate"})
 	switch {
 	case errors.As(err, &dnapi.ErrInvalidCredentials):
 		return nil, InvalidCredentialsError{}
 	case err != nil:
 		return nil, fmt.Errorf("CheckForUpdate error: %s", err)
```
nbrownus: I would update this error to reflect the new api call

nebula/api.go
```diff
@@ -104,18 +104,27 @@ func (c *APIClient) TryUpdate(siteName string, hostID string, privateKey string,
 	// Check for update
 	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
```
nbrownus: Is the anticipation to actually poll or to use the api for its side effects? 30s doesn't match what normal dnclients use for their timeouts but I also don't see this actually maintaining a http request, it fires based on current timers which iirc was 15 minutes or so.

johnmaguire: We will get 409 client gone away messages if the client leaves early. If we plan to use LongPollWait without passing the immediate flag, I believe the intention is to complete the full request, so we should make sure the timeout is longer than 60 seconds, which is the API long poll window.

android/app/src/main/kotlin/net/defined/mobile_nebula/APIClient.kt
```diff
@@ -19,10 +19,10 @@ class APIClient(context: Context) {
         return decodeIncomingSite(res.site)
     }
 
-    fun tryUpdate(siteName: String, hostID: String, privateKey: String, counter: Long, trustedKeys: String): IncomingSite? {
```
johnmaguire: Was there a reason to rename this? tryUpdate was not the name of the API call - it was named this way because it "tried to update" if an update was available. LongPollWait does not update - DoUpdate does. This function does both.

johnmaguire:
I think the most recent commit updated the timeout from 30 seconds to 60 seconds - did you test to see if this was long enough when a LongPollWait doesn't immediately return?

I ask because 60 seconds looks likely to race with the API handler to me, which also uses a 60 second timeout beginning when the request is received (so not including standing up the connection, etc.): https://github.com/DefinedNet/api/blob/b09af0ef4fd2e548d18d61283aef014716b4eb88/internal/app/dnclient_api.go#L577

PR #301, opened 2025-12-04
don't try to enroll with a null code when the screen opens

JackDoan:


johnmaguire (approved):
This looks correct to me...

PR #319, opened 2026-01-21
Fix ipv6 entry in config fields

### LINKED ISSUE #316: Cannot enter IPv6 address for a static host entry nebula IP
## Description
This especially affects IPv6-only nebula networks where it is not possible to enter the nebula IP of a lighthouse.

## Steps to reproduce (Android)
1. Create / Edit site
2. Go to "Lighthouses / Static Hosts"
3. Tap "Add a new entry"
4. Try to enter an IPv6 address into the "Nebula IP" field

My app version is `0.7.0-110 (sha: 891e62f)`

## Extra information
The relevant code seems to be:
- [wrong maximum length of IPAddress fields](https://github.com/DefinedNet/mobile_nebula/blob/891e62f72dc5640e56c6df06b7c2e2f82994dae0/lib/components/IPField.dart#L48) - currently 15, should be 39 (or 41 for the `[2001:0db8:0000:0000:0000:0000:0000:0001]` syntax
- [the IP address validation regex here is wrong](https://github.com/DefinedNet/mobile_nebula/blob/891e62f72dc5640e56c6df06b7c2e2f82994dae0/lib/components/IPField.dart#L59) - I'm not sure what is the intent of the current code, seems to have some extra mobile IME features support, but as far as a list of allowed characters goes it would have to be expanded to something like `[\dabcdefABCDEF:\.,\[\]]+`

nbrownus:
The primary remaining issues are:
- v6 addresses can be long and currently wrap in strange/annoying ways.
- `IPAndPort` style fields we do not display a `[` and `]` around the address section but it also is invalid to input `[` or `]` in the address field.

Fixes #316

### DIFF
--- a/lib/screens/siteConfig/StaticHostsScreen.dart
+++ b/lib/screens/siteConfig/StaticHostsScreen.dart
@@ -86,7 +86,7 @@ class _StaticHostsScreenState extends State<StaticHostsScreen> {
                   color: CupertinoColors.placeholderText.resolveFrom(context),
                 ),
               ),
-              Text(host.nebulaIp),
+              Expanded(child: Text(host.nebulaIp)),
             ],
           ),
           labelWidth: ipWidth,


lib/screens/siteConfig/StaticHostmapScreen.dart
```diff
@@ -89,6 +89,7 @@ class _StaticHostmapScreenState extends State<StaticHostmapScreen> {
                           help: "Required",
                           initialValue: _nebulaIp,
                           ipOnly: true,
+                          enableIPV6: true,
```
JackDoan: is there anywhere with ipv6 not enabled?

JackDoan (approved):

PR #353, opened 2026-02-18
Always on

### LINKED ISSUE #49: Option to always stay connected to Nebula network
I have been using Nebula with my iOS device for quite a while and it has been absolutely great.

I plan to add the iOS devices of some people who are not very tech savvy. My goal is to allow them to access some services via Nebula from any network, but I want it to be totally transparent to them.

I have everything working the way that I want it to at this point, except that if the device restarts for any reason the user needs to manually open the Nebula app or VPN settings to connect.

These iOS devices restart so infrequently that even I often forget for a few minutes that I need to re-connect to Nebula after a restart, so someone who is not very tech savvy is going to find this situation pretty frustrating.

It would be ideal if there were a setting in the Nebula app that allowed a user to just specify that a Nebula should always be running. I don't know much about iOS development so I have no idea whether or not that is straightforward.

If it isn't totally obvious how this could be done, I think it would be helpful to integrate with [Shortcuts ](https://developer.apple.com/documentation/sirikit/offering_actions_in_the_shortcuts_app). I would be satisfied if I could just create a shortcut which ensures that a Nebula network connection is up any time the network configuration changes on the device.

nbrownus:
## Android

Enabling always-on takes 2 steps.
1. Enable for a site in the app. This creates a file which points to the site you want to be always on.
2. Enable always-on in the system settings app for nebula. There is a button that sends you close to the setting below the always-on toggle in the app, we are calling it the John button™.

If the always-on system setting is enabled then always-on doesn't work. We can not reliably detect if its set in a production app.

Small UI change to provide a deep link to the system settings -> network -> vpn and a toggle that sets the armed always on site.
Screenshot 2026-02-19 at 4 12 03 PM

## iOS

Enabling can be done via the UI or system settings (after the UI has saved the site as capable of being on-demand). We can reliably detect if on-demand is enabled.

Each vpn site is its own vpn config in the system settings. It seems the last to set on-demand rules wins or else you must select the active site in the system settings.

**NOTE** Changing any sites always-on status disconnects any existing nebula vpn service. When you manually disconnect a site it also sets always-on to false, otherwise the site would just immediately reconnect.

Small UI change to the site details screen to toggle always-on
Screenshot 2026-02-19 at 12 06 56 PM

Closes #49

### DIFF
--- a/android/app/src/main/kotlin/net/defined/mobile_nebula/MainActivity.kt
+++ b/android/app/src/main/kotlin/net/defined/mobile_nebula/MainActivity.kt
@@ -69,7 +69,6 @@ class MainActivity: FlutterActivity() {
                 "android.registerActiveSite" -> registerActiveSite(result)
                 "android.deviceHasCamera" -> deviceHasCamera(result)
                 "android.openVpnSettings" -> openVpnSettings(result)
-                "android.isAlwaysOnEnabled" -> isAlwaysOnEnabled(result)
 
                 "nebula.parseCerts" -> nebulaParseCerts(call, result)
                 "nebula.generateKeyPair" -> nebulaGenerateKeyPair(result)
@@ -142,11 +141,6 @@ class MainActivity: FlutterActivity() {
         val intent = Intent(Settings.ACTION_VPN_SETTINGS)
         startActivity(intent)
         result.success(null)
-    }
-
-    private fun isAlwaysOnEnabled(result: MethodChannel.Result) {
-        val alwaysOnApp = Settings.Secure.getString(contentResolver, "always_on_vpn_app")
-        result.success(alwaysOnApp == packageName)
     }
 
     private fun nebulaParseCerts(call: MethodCall, result: MethodChannel.Result) {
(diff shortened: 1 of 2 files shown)


ios/Runner/AppDelegate.swift
```diff
@@ -262,10 +262,13 @@ func MissingArgumentError(message: String, details: Any?) -> FlutterError {
...
     #else
-      let manager = self.sites?.getSite(id: id)?.manager
+      let container = self.sites?.getContainer(id: id)
+      let manager = container?.site.manager
+
       manager?.loadFromPreferences { error in
         //TODO: Handle load error
-
+        manager?.isOnDemandEnabled = false
+        manager?.saveToPreferences()
```
jasikpark: saveToPreferences() is asynchronous: https://developer.apple.com/documentation/networkextension/nevpnmanager/savetopreferences(completionhandler:)

```swift
 manager?.saveToPreferences { _ in
    manager?.connection.stopVPNTunnel()
  }
```

We should wait to stop the tunnel until after saving prefs

(recc from `claude /review 353`)

PR #357, opened 2026-02-26
Add excluded apps config to Android

johnmaguire:
Allows users to exclude apps from using the VPN - this is necessary for apps which need to speak to RFC 1918 space, such as IoT devices or cameras, despite the fact that the VPN does not advertise routes for this space.

Uses a 2-phase load for the app icons to avoid an observed slow load of the excluded apps page.

Uninstalled disabled apps are shown at the top, then installed & disabled, then forced disabled apps (the existing hardcoded apps), then all other apps.

image

Tested the behavior with an app that fails without disallowing it.

### DIFF
--- a/android/app/src/main/kotlin/net/defined/mobile_nebula/MainActivity.kt
+++ b/android/app/src/main/kotlin/net/defined/mobile_nebula/MainActivity.kt
@@ -183,7 +183,9 @@ class MainActivity: FlutterActivity() {
 
                 Gson().toJson(apps)
             }
-            result.success(json)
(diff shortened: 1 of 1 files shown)


JackDoan:
Yeah my immediate reaction was "I bet someone is gonna want to only filter some apps sometimes"

nbrownus:
I think for most folks having it on the site is going to be just fine and likely 100% more obvious than if it was hidden in the app settings screen. It's mostly just a bit of a bummer if you have more than 1 site.

PR #361, opened 2026-03-03
Remove IncomingSite, muck with config less in the intermediate layers

nbrownus:
Alternative approach to #358 for config shenanigans 

This adds support for full yaml import.

### DIFF
--- a/android/app/src/main/kotlin/net/defined/mobile_nebula/Sites.kt
+++ b/android/app/src/main/kotlin/net/defined/mobile_nebula/Sites.kt
@@ -331,7 +331,10 @@ class Site(context: Context, siteDir: File) {
         // Parse rawConfig JSON to extract fields needed by VPN service and display
         val rawConfigMap: Map<String, Any?> = try {
             gson.fromJson(rawConfig, object : TypeToken<Map<String, Any?>>() {}.type) ?: emptyMap()
-        } catch (_: Exception) { emptyMap() }
+        } catch (err: Exception) {
+            errors.add("Failed to parse rawConfig: ${err.message}")
+            emptyMap()
+        }
 
         // Parse mtu from rawConfig
         mtu = getConfigInt(rawConfigMap, listOf("tun", "mtu")) ?: 1300
--- a/lib/models/site.dart
+++ b/lib/models/site.dart
@@ -93,6 +93,10 @@ class Site {
     }
   }
 
+  /// Parses site JSON without constructing a full Site (no EventChannel).
+  /// Useful for testing the parse/error logic.
+  static Map<String, dynamic> parseJson(Map<String, dynamic> json) => _fromJson(json);
+
   factory Site.fromJson(Map<String, dynamic> json) {
     var decoded = Site._fromJson(json);
     return Site(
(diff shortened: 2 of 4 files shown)


android/app/src/main/kotlin/net/defined/mobile_nebula/APIClient.kt
```diff
@@ -1,25 +1,21 @@
...
     private val packageInfo = PackageInfo(context)
     private val client = mobileNebula.MobileNebula.newAPIClient(
         "MobileNebula/%s (Android %s)".format(
                 packageInfo.getVersion(),
                 packageInfo.getSystemVersion(),
         ))
-    private val gson = Gson()
 
-    fun enroll(code: String): IncomingSite {
+    fun enroll(code: String): String {
```
JackDoan: hmm..... function does not accomplish much now.....

android/app/src/main/kotlin/net/defined/mobile_nebula/Sites.kt
```diff
@@ -205,98 +205,199 @@ data class DNCredentials(
...
-            if (certs.isEmpty()) {
-                throw IllegalArgumentException("No certificate found")
-            }
-            cert = certs[0]
-            if (!cert!!.validity.valid) {
-                errors.add("Certificate is invalid: ${cert!!.validity.reason}")
+        // Parse rawConfig JSON to extract fields needed by VPN service and display
+        val rawConfigMap: Map<String, Any?> = try {
+            gson.fromJson(rawConfig, object : TypeToken<Map<String, Any?>>() {}.type) ?: emptyMap()
+        } catch (_: Exception) { emptyMap() }
```
JackDoan: just swallowing this error feels like the wrong call

lib/models/site.dart
```diff
@@ -125,93 +98,114 @@ class Site {
...
+    String? key;
+    if (rawConfig['pki'] is Map<String, dynamic>) {
+      final pki = rawConfig['pki'] as Map<String, dynamic>;
+      if (pki['key'] is String) {
+        key = pki['key'] as String;
+      }
+      pki.remove('key');
+    }
+
+    // Parse certs for display via native
```
JackDoan: wait all this is in dart TOO?!

android/app/src/main/kotlin/net/defined/mobile_nebula/MainActivity.kt
```diff
@@ -242,11 +239,17 @@ class MainActivity: FlutterActivity() {
             return result.error("failure", "Site config was incomplete, please review and try again", null)
         }
 
-        if (site.alwaysOn == true) {
-            if (activeSiteId != null && activeSiteId != site.id) {
-                stopSite { sites?.getSite(site.id)?.let { startSiteDirectly(it) } }
-            } else if (activeSiteId != site.id) {
-                sites?.getSite(site.id)?.let { startSiteDirectly(it) }
+        // Check if we need to start/restart for always-on
+        val gson = Gson()
```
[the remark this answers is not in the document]
nbrownus: Only 1 site can actually be the always on site, in practice what happens is toggling a new site always on will start it and turn always on off on the old site.

The other side of this is in loading a site, it technically has the boolean in the model saved to disk but ultimately matches what the always on file says is the always on site.

JackDoan (dismissed):

PR #388, opened 2026-07-23
Use a better file picker

nbrownus:
Trade `file_picker` for `file_selector` and drop temporary files

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -59,7 +59,6 @@ class AppState extends State<App> {
 
   @override
   void initState() {
-    // Waits until here so the binding is up, main runs before anyone has made one
     unawaited(Utils.clearLegacyPickedFiles());
 
     //TODO: wait until settings is ready?
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -59,6 +59,9 @@ class AppState extends State<App> {
 
   @override
   void initState() {
+    // Waits until here so the binding is up, main runs before anyone has made one
+    unawaited(Utils.clearLegacyPickedFiles());
+
     //TODO: wait until settings is ready?
     settings.onChange().listen((_) {
       setState(() {
--- a/lib/services/utils.dart
+++ b/lib/services/utils.dart
@@ -2,10 +2,13 @@ import 'dart:io';
 
 import 'package:file_selector/file_selector.dart';
 import 'package:flutter/material.dart';
+import 'package:logging/logging.dart';
 import 'package:mobile_nebula/main.dart';
 import 'package:path/path.dart' as p;
 import 'package:path_provider/path_provider.dart';
 import 'package:url_launcher/url_launcher_string.dart';
+
+final _log = Logger('utils');
 
 class Utils {
   /// Minimum size (width or height) of a interactive component
@@ -128,7 +131,7 @@ class Utils {
     }
   }
 
-  static Future<String?> pickFile(BuildContext context) async {
+  static Future<String?> pickFile() async {
     final file = await openFile();
     if (file == null) {
       return null;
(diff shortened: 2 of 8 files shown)


lib/services/utils.dart
```diff
@@ -127,14 +129,52 @@ class Utils {
   }
 
   static Future<String?> pickFile(BuildContext context) async {
```
johnmaguire: nit: `context` is unused now

lib/services/utils.dart
```diff
@@ -127,14 +129,52 @@ class Utils {
...
       return null;
     }
 
-    final file = File(result.files.first.path!);
-    return file.readAsString();
+    final content = await file.readAsString();
+
+    // We get a copy, not the original, so don't leave a plaintext key sitting around
+    try {
+      await File(file.path).delete();
```
johnmaguire: Android drops each copy in its own `<cacheDir>/<uuid>/`, and the plugin's `targetDirectory.deleteOnExit()` (`FileUtils.java:130`) never fires there - the javadoc attempts deletion only on *normal* VM termination, which an Android process kill is not. Worth adding a non-recursive `.parent.delete()` here; it throws harmlessly into the existing catch on iOS, where the parent is shared.

```suggestion
      await File(file.path).delete();
      // Android drops each copy in its own <cacheDir>/<uuid>/; fails harmlessly
      // on iOS where the parent is shared
      await File(file.path).parent.delete();
```

lib/services/utils.dart
```diff
@@ -127,14 +129,52 @@ class Utils {
...
+    } catch (err) {
+      // Ignoring file delete errors
```
johnmaguire: Possible plaintext key material stayed on disk, maybe worth a log at least

lib/services/utils.dart
```diff
@@ -127,14 +129,52 @@ class Utils {
...
+    return content;
+  }
+
+  /// Deletes copies left behind by file_picker, which we used to use. It only cleared
+  /// its cache on the next pick, so the last thing anyone imported is still in there.
+  static Future<void> clearLegacyPickedFiles() async {
+    try {
+      final cacheDir = await getTemporaryDirectory();
+      final legacyCache = Directory(p.join(cacheDir.path, 'file_picker'));
+      if (legacyCache.existsSync()) {
```
johnmaguire: `existsSync` / `deleteSync` / `listSync` block the isolate during startup for no benefit, since this is already fire-and-forget. The dart:io docs say it directly at `file_system_entity.dart:230`: "prefer the asynchronous version to avoid blocking your program."

```suggestion
      if (await legacyCache.exists()) {
        await legacyCache.delete(recursive: true);
      }
```

lib/services/utils.dart
```diff
@@ -127,14 +129,52 @@ class Utils {
...
+        legacyCache.deleteSync(recursive: true);
+      }
+    } catch (err) {
+      // Ignoring cleanup errors
+    }
+
+    // iOS document imports land in <tmp>/<bundle id>-Inbox. systemTemp because
```
johnmaguire: I could not confirm anywhere that `.import` stages into `<tmp>/<bundle id>-Inbox`. Apple's picker docs do not say, and the archived File System guide's `Documents/Inbox` entry describes a different API (UIDocumentInteractionController and Mail attachments, not the picker). Worth a one-line `debugPrint` of the path `openFile()` returns on a device before merging, because if it is not under `systemTemp` then the whole iOS branch below matches nothing and is dead code.

lib/services/utils.dart
```diff
@@ -127,14 +129,52 @@ class Utils {
...
+    // path_provider points getTemporaryDirectory at Library/Caches, not tmp, and it
+    // is only our sandbox on iOS. Inboxes only, Share is using the rest of tmp.
```
johnmaguire: I think this premise is inverted. `Share` writes to `getTemporaryDirectory()` (`lib/services/share.dart:24`), which path_provider_foundation maps to `NSCachesDirectory` on iOS - `path_provider_foundation_real.dart:41`, pinned at 2.6.0 in the lockfile. Share never touches tmp, so there is nothing of ours here to protect from a wider sweep.

lib/services/utils.dart
```diff
@@ -127,14 +129,52 @@ class Utils {
...
+    if (!Platform.isIOS) {
+      return;
+    }
+
+    try {
+      for (final entry in Directory.systemTemp.listSync()) {
+        if (entry is Directory && entry.path.endsWith('-Inbox')) {
```
johnmaguire: Blocking, I think. file_picker 12 does not hand back the `-Inbox` path - it copies to a loose file at the top of tmp. `IOSFilePickerHandler.swift:208` calls `copyToTemporaryDirectory`, defined at `:334-345`, which writes to `NSTemporaryDirectory()/<filename>`. So this filter walks straight past the plaintext key it exists to delete.

That file is exactly what the old `FilePicker.clearTemporaryFiles()` used to remove, since it wiped every entry in tmp (`:308-320`) - which means after this upgrade the last-imported key stays on disk permanently, the failure mode the method is named for.

```suggestion
        if (entry is File || (entry is Directory && entry.path.endsWith('-Inbox'))) {
          entry.deleteSync(recursive: true);
        }
```

pubspec.yaml
```diff
@@ -23,16 +23,15 @@ dependencies:
     sdk: flutter
 
   path_provider: ^2.0.11
-  # 12.0.0 drops the DKImagePickerController dependency chain on iOS in favor of
-  # PHPickerViewController, keeping photo library permissions out of our binary
-  file_picker: ^12.0.0-beta.7
+  # We only pick documents (certs, keys, configs). file_selector is first party and
+  # has no media picker code, so photo library permissions can never sneak into our binary
+  file_selector: ^1.1.0
```
johnmaguire, line 28: FYI: the allowed-type root narrows from `public.item` (file_picker `IOSFilePickerHandler.swift:67`) to `public.data` (`file_selector_ios.dart:54`, pinned 0.5.3+5). Should be invisible for certs, keys and YAML, but it is a real behavior change. Claude suggests testing the Files-app-via-iCloud at least.

lib/main.dart
```diff
@@ -59,6 +59,9 @@ class AppState extends State<App> {
 
   @override
   void initState() {
+    // Waits until here so the binding is up, main runs before anyone has made one
```
johnmaguire: I don't quite follow the comment, fwiw

pubspec.yaml
nbrownus: Tested icloud and it works fine 👍

lib/services/utils.dart
```diff
@@ -126,15 +131,60 @@ class Utils {
...
-    if (result == null) {
+  static Future<String?> pickFile() async {
+    final file = await openFile();
+    if (file == null) {
       return null;
     }
 
-    final file = File(result.files.first.path!);
-    return file.readAsString();
+    final content = await file.readAsString();
```
johnmaguire, line 140: Blocking: decode the selected bytes as UTF-8 explicitly here. The locked file_selector_android 0.5.2+8 constructs the result with XFile.fromData(file.bytes, ...): [file_selector_android source](https://github.com/flutter/packages/blob/a89dc58e2bc2d511a49249ed90039df7ebc71ec2/packages/file_selector/file_selector_android/lib/src/file_selector_android.dart#L59-L71). The locked cross_file 0.3.5+4 handles that byte-backed readAsString call with String.fromCharCodes(_bytes), not the requested UTF-8 decoder: [cross_file source](https://github.com/flutter/packages/blob/b967797f07f1e4fc4fb264f5c8e5181b1385d9d9/packages/cross_file/lib/src/types/io.dart#L113-L120). UTF-8 bytes [0xC3, 0xA9] therefore become U+00C3 U+00A9 instead of U+00E9, silently corrupting non-ASCII YAML on Android. Dart documents that [String.fromCharCodes treats its input as character codes](https://api.dart.dev/dart-core/String/String.fromCharCodes.html), while [Utf8Decoder converts UTF-8 bytes to text](https://api.dart.dev/dart-convert/Utf8Decoder-class.html). Please import dart:convert and use utf8.decode(await file.readAsBytes()).

lib/services/utils.dart
```diff
@@ -126,15 +131,60 @@ class Utils {
...
+
+    // It is a copy, so don't leave a plaintext key around. Android puts each in its own
+    // <cacheDir>/<uuid>/, so drop the empty parent too; on iOS that throws into the catch.
+    try {
```
johnmaguire, line 144: Cleanup gap: the read on line 140 occurs before this try block, so a read or decode error skips deletion. On iOS, the pinned plugin creates UIDocumentPickerViewController in import mode and returns its path as a path-backed XFile: [native source](https://github.com/flutter/packages/blob/03299cdac3c97877daee7583057ee45ff6372ece/packages/file_selector/file_selector_ios/ios/file_selector_ios/Sources/file_selector_ios/FileSelectorPlugin.swift#L61-L69) and [Dart source](https://github.com/flutter/packages/blob/03299cdac3c97877daee7583057ee45ff6372ece/packages/file_selector/file_selector_ios/lib/file_selector_ios.dart#L24-L35). [Apple documents](https://developer.apple.com/library/archive/documentation/FileManagement/Conceptual/DocumentPickerProgrammingGuide/AccessingDocuments/AccessingDocuments.html) that import mode supplies a temporary copy in the app sandbox that remains available until the app closes. A path-backed XFile delegates readAsString to File.readAsString, whose default UTF-8 decoder rejects malformed input: [cross_file source](https://github.com/flutter/packages/blob/b967797f07f1e4fc4fb264f5c8e5181b1385d9d9/packages/cross_file/lib/src/types/io.dart#L113-L120) and [Utf8Decoder docs](https://api.dart.dev/dart-convert/Utf8Decoder/Utf8Decoder.html). Please wrap the read and return in an outer try/finally, retaining the inner try/catch for best-effort deletion; [Dart documents finally](https://dart.dev/language/error-handling#finally) as the mechanism for cleanup that must run after errors.

PR #389, opened 2026-07-24
Truncate long names instead of running off screen

johnmaguire:
Long names could run past the right edge of the screen. Two places lay a
name out in a `Row` with no width constraint, so the text overlapped the
row's chevron and continued off screen:

- **Active/Pending Tunnels** - the cert name (`site_tunnels_screen.dart`)
- **Firewall Rule** - a group name chip (`pill_chip.dart`)

Both now ellipsize at the edge.

## Screenshots

360dp wide, the common Android width. The red bar and hazard stripes are
Flutter's debug overflow indicator.

### Active Tunnels

| before | after |
| --- | --- |
| tunnels-before | tunnels-after |

### Firewall Rule groups

| before | after |
| --- | --- |
| image | image |


johnmaguire:
Updated to wrap group names
image

nbrownus (approved):

PR #390, opened 2026-07-29
Drop fastlane match, use key material that live in aws secrets manager instead

nbrownus:
This moves the apple signing key material to aws secrets manager instead of each repo/org having its own copy.


.github/workflows/release.yml
```diff
@@ -156,14 +120,105 @@ jobs:
...
+          p12-file-base64: ${{ env.SIGNING_P12_BASE64 }}
+          p12-password: ${{ env.SIGNING_PASSWORD }}
+
+      # The action imports but does not check the chain validates, which is how a p12
+      # missing its intermediate reaches a failing codesign
+      - name: Check the identity is usable
+        run: |
+          identities=$(security find-identity -v -p codesigning signing_temp.keychain)
+          case "$identities" in
+          *"$SIGNING_IDENTITY_SHA1"*) ;;
```
johnmaguire, line 157: If `SIGNING_IDENTITY_SHA1` is empty this becomes `**` which matches any output.

.github/workflows/release.yml
```diff
@@ -156,14 +120,105 @@ jobs:
...
+      # Both platforms are built by now and neither publish has happened, so a
+      # failure above this line still publishes nothing anywhere. Keep these two
+      # adjacent and keep them last.
+      - name: Publish to iOS TestFlight
         run: |
           cd ios
           fastlane release
 
+      # Nothing below needs them, and the Sentry upload runs code we do not own
+      - name: Drop the Apple credentials
```
johnmaguire, line 214: Claude says that even after this step, the signing private key imported by `import-codesign-certs` is in `signing_temp.keychain` still. It recommends `security delete-keychain signing_temp.keychain` to ensure Sentry, etc. cannot touch them.

ios/fastlane/Fastfile
```diff
@@ -65,19 +49,30 @@ platform :ios do
         version_number: ENV['BUILD_NAME']
     )
 
+    # Archiving and exporting each resolve provisioning, so both need the key.
+    # fetch rather than [] so a missing one fails here, not inside xcodebuild.
+    asc = [
+      "-allowProvisioningUpdates",
+      "-authenticationKeyPath #{ENV.fetch('APP_STORE_CONNECT_API_KEY_KEY_FILEPATH')}",
```
johnmaguire: nit: unquoted path

.github/workflows/release.yml
```diff
@@ -156,14 +120,105 @@ jobs:
...
+          *) printf '%s\n' "$identities" >&2; exit 1 ;;
+          esac
+
+      # Under the names fastlane's app_store_connect_api_key and xcodebuild's
+      # -authenticationKeyPath expect. Replaces the AC_API_KEY_* GitHub secrets.
+      - name: Write the App Store Connect key
+        run: |
+          key_path="$RUNNER_TEMP/AuthKey_${ASC_KEY_ID}.p8"
+          printf '%s\n' "$ASC_PRIVATE_KEY" > "$key_path"
+          chmod 600 "$key_path"
```
johnmaguire: nit: safer to pre-create the directory set to 600

johnmaguire (commented):
nit: several comments narrate the migration rather than current behavior — "Replaces the AC_API_KEY_* GitHub secrets", "there is no longer a match repository", "What changed is the signing style". Once merged these read as noise; the surrounding "why" content (keychain competition, ordering rationale) is the good part and can stand alone.

johnmaguire (approved):

PR #391, opened 2026-07-30
Fix the release for iOS

nbrownus:
Moves more out of fastlane (I'm trying to deprecate it) and fixes some bugs introduced in #390

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -229,7 +229,7 @@ jobs:
           for app in Payload/Runner.app Payload/Runner.app/PlugIns/*.appex; do
             [ -e "$app" ] || continue
             rm -f codesign[0-9]
-            codesign -d --extract-certificates "$app" 2>/dev/null
+            codesign -d --extract-certificates "$app"
             signer=$(openssl x509 -inform DER -in codesign0 -noout -fingerprint -sha1 |
               cut -d= -f2 | tr -d ':')
             if [ "$signer" != "$SIGNING_IDENTITY_SHA1" ]; then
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -210,17 +259,22 @@ jobs:
...
-          {
-            echo "SIGNING_P12_BASE64="
-            echo "SIGNING_PASSWORD="
-            echo "ASC_PRIVATE_KEY="
-            echo "APP_STORE_CONNECT_API_KEY_KEY_FILEPATH="
-          } >> "$GITHUB_ENV"
+          # SECRETS_LIST_CLEAN_UP is the action's own list, so this does not rot
+          python3 -c '
+          import json, os
+          names = json.loads(os.environ.get("SECRETS_LIST_CLEAN_UP") or "[]")
```
johnmaguire: I assume it's intentional that this fails silently if `SECRETS_LIST_CLEAN_UP` does not exist.

.github/workflows/release.yml
```diff
@@ -190,6 +224,21 @@ jobs:
           cp ios/MobileNebula.ipa build/app/test-ios
           cd build/app/test-ios
           unzip MobileNebula.ipa
+
+          # Importing a certificate does not constrain what Xcode picks
+          for app in Payload/Runner.app Payload/Runner.app/PlugIns/*.appex; do
+            [ -e "$app" ] || continue
+            rm -f codesign[0-9]
+            codesign -d --extract-certificates "$app" 2>/dev/null
```
johnmaguire: Won't be able to see what the failure was with `2>/dev/null` if this fails.

johnmaguire (dismissed):
LGTM

johnmaguire (approved):

PR #392, opened 2026-07-31
Use network name for mobile enrollments

johnmaguire:
Matches DNClient behavior.

### DIFF
--- a/lib/screens/main_screen.dart
+++ b/lib/screens/main_screen.dart
@@ -293,7 +293,7 @@ class MainScreenState extends State<MainScreen> {
       ),
       ListTile(
         title: Text('Enroll with defined.net'),
-        subtitle: Text('Join your organizations network'),
+        subtitle: Text('Join a Managed Nebula network'),
         trailing: arrowIcon,
         onTap: () {
           // Remove the modal
--- a/nebula/api.go
+++ b/nebula/api.go
@@ -66,7 +66,14 @@ func (c *APIClient) Enroll(code string) (*EnrollResult, error) {
 		return nil, fmt.Errorf("unexpected failure: %s", err)
 	}
 
-	site, err := newDNSite(meta.Org.Name, cfg, string(pkey), *creds)
+	// Name the site after the network to match DNClient. Fall back to the org
+	// name in case the API returns an empty network name.
+	name := meta.Network.Name
+	if name == "" {
+		name = meta.Org.Name
+	}
+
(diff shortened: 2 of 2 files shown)


lib/screens/main_screen.dart
```diff
@@ -293,7 +293,7 @@ class MainScreenState extends State<MainScreen> {
       ),
       ListTile(
         title: Text('Enroll with defined.net'),
-        subtitle: Text('Join your organizations network'),
```
johnmaguire, line 296: kinda meh on this one, but it matches our naming elsewhere. alternatively just `s/organizations/organization's/`

jasikpark (approved):
**praise:** Verified the one sharp edge — at the pinned dnapi commit (`c66f625d8dfd`), `ConfigMeta.Network` is a value struct populated on every successful enroll, so `meta.Network.Name` is always safe to read and the empty-string fallback covers older responses cleanly.

**note (non-blocking):** Network renames still won't propagate — `doUpdate`/`doConfigUpdate` discard the returned `meta` and reuse the caller's `siteName`, so a site keeps its enroll-time name forever. That was already true with org names, but networks get renamed much more often than orgs, so the staleness is more likely to actually be noticed now. A follow-up could use the meta returned by `DoUpdate`/`DoConfigUpdate` to refresh the site name. (Existing enrolled sites also keep their org-derived names — only new enrollments pick up the network name.)

**note (non-blocking):** Orgs that never renamed their default network will now enroll mobile sites literally named "Network1", which is less informative than the org name it replaces. It matches DNClient, so consistency wins — but it nudges up the priority of [DefinedNet/webclient#4055](https://github.com/DefinedNet/webclient/issues/4055) (default network name is user-visible).

— Caleb + Cache

johnmaguire:
> note (non-blocking): Network renames still won't propagate — doUpdate/doConfigUpdate discard the returned meta and reuse the caller's siteName, so a site keeps its enroll-time name forever. That was already true with org names, but networks get renamed much more often than orgs, so the staleness is more likely to actually be noticed now. A follow-up could use the meta returned by DoUpdate/DoConfigUpdate to refresh the site name. (Existing enrolled sites also keep their org-derived names — only new enrollments pick up the network name.)

This was intentional - we allow mobile users to rename the site in the list, and don't want to overwrite their rename. We could invent more complicated logic, but that's out of scope of this PR.

nbrownus (approved):

PR #393, opened 2026-07-31
Indicate relayed tunnels and fix a build problem

nbrownus:
Adds some basic relay information for tunnels using a relay

IMG_3963
IMG_3964

Also ran into a build issue from our ios target getting moved back to 13, this asserts we are at the correct version (14)

### DIFF
--- a/ensure-mobile-nebula.sh
+++ b/ensure-mobile-nebula.sh
@@ -14,6 +14,8 @@ FRAMEWORK="$BINARIES/MobileNebula.xcframework"
 MARKER="$BINARIES/MobileNebula.freshness"
 STAMP="$BINARIES/MobileNebula.stamp"
 BUILD_LOG="$BINARIES/MobileNebula.build.log"
+FLUTTER_LOG="$BINARIES/flutter-config.log"
+MANIFEST="$ROOT/ios/Flutter/ephemeral/Packages/FlutterGeneratedPluginSwiftPackage/Package.swift"
 
 mkdir -p "$BINARIES"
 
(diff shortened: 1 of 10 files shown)


ensure-mobile-nebula.sh
```diff
@@ -35,6 +37,38 @@ unset SDKROOT MACOSX_DEPLOYMENT_TARGET IPHONEOS_DEPLOYMENT_TARGET \
...
+    [ -f "$MANIFEST" ] || return 0
+    sed -n 's/.*\.iOS("\([0-9.]*\)").*/\1/p' "$MANIFEST" | head -1
+}
+
+TARGET="$(project_ios_target)"
+HAVE="$(manifest_ios_target)"
+
+if [ -n "$TARGET" ] && [ "$HAVE" != "$TARGET" ]; then
+    echo "FlutterGeneratedPluginSwiftPackage is iOS ${HAVE:-absent}, project is $TARGET, running flutter config"
+    if flutter build ios --config-only > "$FLUTTER_LOG" 2>&1; then
```
jasikpark: issue: `flutter build ios --config-only` has to run from the Flutter project root, but this script's primary caller is the Runner scheme's pre-build action, which doesn't run at the repo root — `gen-artifacts.sh` calls out this exact case ("Callers are not always at the repo root, Xcode pre-actions in particular") and `cd`s to compensate. As written, under the pre-action flutter errors out, we take the warning branch (buried in `flutter-config.log`, which Xcode won't surface), and the SPM rejection this exists to prevent still happens. It works by hand from the root and in CI, which is probably why it looked fixed.

```suggestion
    if (cd "$ROOT" && flutter build ios --config-only > "$FLUTTER_LOG" 2>&1); then
```

— Caleb + Cache

ensure-mobile-nebula.sh
```diff
@@ -35,6 +37,38 @@ unset SDKROOT MACOSX_DEPLOYMENT_TARGET IPHONEOS_DEPLOYMENT_TARGET \
...
+
+manifest_ios_target() {
+    [ -f "$MANIFEST" ] || return 0
+    sed -n 's/.*\.iOS("\([0-9.]*\)").*/\1/p' "$MANIFEST" | head -1
+}
+
+TARGET="$(project_ios_target)"
+HAVE="$(manifest_ios_target)"
+
+if [ -n "$TARGET" ] && [ "$HAVE" != "$TARGET" ]; then
```
jasikpark, line 67: nit (non-blocking): two harmless edge cases in this compare. A format drift between the pbxproj value and what flutter writes (`14` vs `14.0`) would make it warn and re-run flutter config on every build; and if the pbxproj ever holds multiple distinct deployment targets, `sort -u | head -1` picks the lexicographic min rather than the numeric one. Both fine today — just noting in case the every-build flap ever shows up.

— Caleb + Cache

nebula/go.mod
```diff
@@ -42,7 +42,7 @@ require (
 	github.com/vishvananda/netns v0.0.5 // indirect
 	go.yaml.in/yaml/v2 v2.4.3 // indirect
 	go.yaml.in/yaml/v3 v3.0.4 // indirect
-	golang.org/x/mobile v0.0.0-20260709172247-6129f5bee9d5 // indirect
+	golang.org/x/mobile v0.0.0-20260730202154-c700fe717e6e // indirect
```
jasikpark, line 45: nit (non-blocking): go.sum still carries the old `20260709` x/mobile entries alongside the new ones — `go mod tidy` in `nebula/` would drop them.

— Caleb + Cache

jasikpark (commented):
Reviewed with most of the attention on the Dart↔Go contract, since none of the Go in this repo changes.

**Verified:**
- The JSON keys match nebula v1.11.0's `ControlHostInfo` tags exactly (`currentRelaysToMe`/`currentRelaysThroughMe`, `[]netip.Addr` → string arrays), so the already-pinned version emits everything the model parses — no Go-side changes needed, as expected.
- `isRelayed` semantics check out against `copyHostInfo`: relay IPs persist in relay state after a direct path exists, and a relayed tunnel's `CurrentRemote` is the zero `AddrPort`, which marshals to `""` → `null` on the Dart side. Gating on `currentRemote == null && currentRelaysToMe.isNotEmpty` is exactly right.

praise: the "has a relay but is also direct → not relayed" test pins the tricky quadrant of that truth table — nice.

One real issue inline (the `flutter build ios --config-only` call is cwd-dependent and pre-build actions don't run at the repo root); the rest are non-blocking nits.

— Caleb + Cache

nebula/go.mod
nbrownus: And the build puts them back, we always install the latest gomobile.

jasikpark (approved):
All addressed — the `cd "$ROOT"` subshell fixes the pre-action case, and the version compare is now numeric + normalized. Withdrawing the go.sum nit per your explanation (verified: nothing imports x/mobile, so the Makefile's unpinned `go get` re-adds it at latest every build — tidy would just be churn).

— Caleb + Cache

nebula/go.mod
jasikpark: You're right — withdrawn. Nothing here imports x/mobile, so `tidy` would actually drop the requirement entirely and the Makefile's unpinned `go get` re-adds it at latest on the next build. Working as intended; resolving.

— Caleb + Cache

PR #394, opened 2026-08-03
Allow local dns resolver config for managed sites

nbrownus:
Also adds some stubs to allow a dn admin from locking out local configuration, this feature does not exist in dn yet.

### DIFF
--- a/ios/NebulaNetworkExtension/Site.swift
+++ b/ios/NebulaNetworkExtension/Site.swift
@@ -159,7 +159,8 @@ class UnsafeRoute: Codable {
 
 /// Saves a site JSON string to disk. Extracts key and dnCredentials into
 /// encrypted storage and writes the remaining config to config.json.
-/// If existingSite is provided, client-only fields (sortKey) will be preserved from it.
+/// If existingSite is provided, client-only fields (sortKey, lastManagedUpdate, dnsResolvers and
+/// matchDomains) will be preserved from it.
 func saveSiteToDisk(jsonString: String, existingSite: Site? = nil) throws {
   guard let jsonData = jsonString.data(using: .utf8),
     let obj = try? JSONSerialization.jsonObject(with: jsonData),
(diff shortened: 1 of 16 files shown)


ios/NebulaNetworkExtension/Site.swift
```diff
@@ -448,19 +480,22 @@ class Site: Encodable {
...
+    managedMatchDomains = mobileNebulaConfig?["match_domains"] as? [String] ?? []
 
-    if let domains = mobileNebulaConfig?["match_domains"] as? [String] {
-      matchDomains = domains
-    } else {
-      matchDomains = []
-    }
+    // Absent has to mean allowed. Every config written before this shipped lacks the flag, and so
+    // does every DN backend until the server side lands, so failing closed would lock dns
+    // resolvers on every managed site the moment someone takes the app update.
```
jasikpark, line 497: **question (non-blocking):** The API's `definednet.dns` config includes `search_domains` alongside `resolver_addrs` and `match_domains`, but only resolvers and match domains get the local/managed split here. Is `search_domains` deliberately out of scope for mobile (iOS conflates search and match via `NEDNSSettings`), or should it get the same treatment?

lib/screens/siteConfig/site_config_screen.dart
```diff
@@ -405,16 +405,27 @@ class SiteConfigScreenState extends State<SiteConfigScreen> {
...
-                    site.mtu = settings.mtu;
+
+                    // A managed site only gets to change the client side settings below, writing
+                    // the rest back would bake our defaults into the config DN handed us.
+                    if (!site.managed) {
+                      site.cipher = settings.cipher;
+                      site.lhDuration = settings.lhDuration;
+                      site.port = settings.port;
+                      site.logVerbosity = settings.verbosity;
+                      site.unsafeRoutes = settings.unsafeRoutes;
```
jasikpark, line 416: **praise:** This quietly fixes a real bug — before this, saving advanced settings on a managed site wrote all values back unconditionally, baking defaults into the admin-supplied config. Good catch.

jasikpark (commented):
Clean architecture — moving the migration chain into Go so Kotlin/Swift are thin wrappers means v3+ won't need native changes. The three-tier DNS resolution (admin-locked > local override > admin default) with null-vs-empty semantics is the right model, and the test coverage is thorough on both the Go and Dart sides.

The verbatim-return optimization for `MigrateConfig` (byte-identical when no migration needed, letting callers skip the disk write on a string comparison) is a nice detail.

— Caleb + Cache

PR #395, opened 2026-08-03
Fix poisoned Go cache in CI

johnmaguire:


jasikpark (approved):
Faithful extension of api#2220's Go-cache fix to this repo's shape (4 workflows, 5 setup-go call sites). Verified: distinct cache-key prefixes per job (gofmt-go-/gotest-go-/smoke-android-go-/smoke-ios-go-, no collisions), OS-correct build-cache paths (~/.cache/go-build on ubuntu, ~/Library/Caches/go-build on macos runners), and release.yml deliberately drops to no-cache (cache: false, no restore) rather than racing the shared key — reasonable given it only runs on version-tag pushes. fluttercheck.yml/swiftfmt.yml correctly left untouched, neither uses setup-go. CI green across gofmt, Go tests, and both smoke builds.

PR #401, opened 2026-08-31
Resolve site DNS settings through a shared Go helper

johnmaguire:
## What

Moves site DNS resolution into a single Go helper, `EffectiveDNS`, consumed by both platform layers instead of each parsing `rawConfig` on its own. There are exactly two DNS channels:

1. `dnsOverride` - a device-local, client-only site field (preserved across managed config updates, like `excludedApps`/`sortKey`). The explicit `enabled` flag lets a user disable managed DNS on a device without substituting their own resolvers. This is now the only channel for user-configured DNS, on managed and unmanaged sites alike.
2. Managed `definednet.dns` from `rawConfig` - non-empty `resolver_addrs` is the install signal, matching dnclient. Applies only while the override is not enabled.

The helper returns `{resolvers, matchDomains, searchDomains, source}` as JSON; empty lists marshal as `[]` for the platform parsers.

## Migration

The legacy manual channel (`mobile_nebula.dns_resolvers` / `match_domains` / `search_domains` inside `rawConfig`) is dropped. A `configVersion` 2 migration (`MigrateConfigV2` in Go, run by both platform `ConfigMigrator`s on site load) hoists those settings into `dnsOverride` with `enabled: true`, removes the keys, and deletes the `mobile_nebula` block if empty. An existing `dnsOverride` is never clobbered. The v0 to v1 migration still emits `mobile_nebula.dns_resolvers`, so v0 sites migrate through the chain. `saveSite` on both platforms now stamps `configVersion: 2`. YAML imports (`Site.fromYaml`) apply the same hoist, since imported configs never pass through `ConfigMigrator` and are stamped at the current version on save.

## Platform changes

- **Android**: `Site` resolves `dnsResolvers`/`matchDomains`/`searchDomains` via the helper; `NebulaVpnService` applies search domains with `addSearchDomain` and logs a warning when match domains are set, since `VpnService` cannot split DNS by domain. `dnsOverride` is exported to Flutter.
- **iOS**: `Site` resolves through the same helper; `PacketTunnelProvider` now applies `searchDomains` to `NEDNSSettings`. `dnsOverride` is a typed `Codable` struct so it rides along in the `Encodable` output to Flutter.
- **Flutter**: `Site` round-trips `dnsOverride` through save (omitted when unset so the platform preserves any existing override). The `dnsResolvers`/`matchDomains` getters and setters read and write the override instead of `rawConfig`; writing any DNS setting sets `enabled: true`.

## Testing

- `go test ./...` in `nebula/` covers channel precedence, the enabled-but-empty override (deliberately disables DNS), the inert empty managed block, `mobile_nebula.*` being ignored post-migration, and the v2 migration itself (hoist, key removal, preservation of unrelated `mobile_nebula` keys and existing overrides).
- `flutter test` covers the new Dart accessors and save JSON behavior.
- Builds against upstream `slackhq/nebula v1.11.0` with no go.mod changes.

### DIFF
--- a/lib/models/site.dart
+++ b/lib/models/site.dart
@@ -47,6 +47,9 @@ class Site {
   late bool alwaysOn;
   late List<String> excludedApps;
 
+  // Device-local DNS override (client-only field, preserved across managed config updates)
+  Map<String, dynamic>? dnsOverride;
+
   // DN management
   late bool managed;
   late DateTime? lastManagedUpdate;
@@ -70,6 +73,7 @@ class Site {
     this.lastManagedUpdate,
     this.alwaysOn = false,
     List<String>? excludedApps,
+    this.dnsOverride,
   }) {
     this.id = id ?? uuid.v4();
     this.rawConfig = rawConfig ?? {};
@@ -119,6 +123,7 @@ class Site {
       lastManagedUpdate: decoded['lastManagedUpdate'],
       alwaysOn: decoded['alwaysOn'],
       excludedApps: decoded['excludedApps'],
+      dnsOverride: decoded['dnsOverride'],
     );
   }
 
@@ -206,6 +211,7 @@ class Site {
     lastManagedUpdate = decoded['lastManagedUpdate'];
     alwaysOn = decoded['alwaysOn'];
     excludedApps = decoded['excludedApps'];
+    dnsOverride = decoded['dnsOverride'];
   }
 
   static Map<String, dynamic> _fromJson(Map<String, dynamic> json) {
@@ -259,6 +265,7 @@ class Site {
       "lastManagedUpdate": json["lastManagedUpdate"] == null ? null : DateTime.parse(json["lastManagedUpdate"]),
       "alwaysOn": json['alwaysOn'] ?? false,
       "excludedApps": excludedApps,
+      "dnsOverride": json['dnsOverride'] == null ? null : Map<String, dynamic>.from(json['dnsOverride']),
     };
   }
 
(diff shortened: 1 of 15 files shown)


jasikpark (dismissed):
Clean centralization — routing DNS precedence through a single Go helper instead of per-platform parsing is the right call. The migration chain (v0→v1→v2, with YAML import covering the ConfigMigrator bypass) is thorough, and the `enabled`-flag design for the override is a good semantic choice over "non-empty resolvers = enabled."

The empty-write guard in commit 5 is the kind of defensive detail that saves a real user from a silent DNS-disable — glad to see it tested explicitly.

— Caleb + Cache

lib/models/site.dart
```diff
@@ -259,6 +306,11 @@ class Site {
       "lastManagedUpdate": json["lastManagedUpdate"] == null ? null : DateTime.parse(json["lastManagedUpdate"]),
       "alwaysOn": json['alwaysOn'] ?? false,
       "excludedApps": excludedApps,
+      "dnsOverride": json['dnsOverride'] == null ? null : Map<String, dynamic>.from(json['dnsOverride']),
+      // Effective DNS resolved by the platform (keyed dnsResolvers/matchDomains
+      // in the platform Site serialization)
+      "effectiveDnsResolvers": stringList(json['dnsResolvers']),
+      "effectiveMatchDomains": stringList(json['matchDomains']),
```
jasikpark, line 313: **question (non-blocking):** Both platforms serialize `searchDomains` to Flutter, but there's no `effectiveSearchDomains` parsed here and no UI surface for it. Is this deliberate — the DNS resolvers screen doesn't have a search domains section, so they're intentionally platform-only for now?

— Caleb + Cache

nbrownus (approved):

PR #404, opened 2026-09-22
Update nebula to v1.11.2

jasikpark:
Bumps the Go bridge from nebula v1.11.0 to v1.11.2. Two files, `nebula/go.mod` and `nebula/go.sum`.

v1.11.2 is the v1.11 line's second patch: Windows NLMTU for IPv6 (#1871), a transport checksum recompute on self-forwarded packets (#1862), a Windows MTU smoke test (#1873), and underlay address advertisement through the new `nicfinder` package (#1881). `nicfinder` is what adds `DefinedNet/netlink` to the module graph.

No code changes: the logrus -> slog migration already landed in #387.

Verified in `nebula/`: `go build ./...`, `GOOS=android GOARCH=arm64 go build ./...`, `GOOS=ios GOARCH=arm64 CGO_ENABLED=1 go build ./...`, `go vet ./...`, `go test ./...`. The aar and xcframework were not rebuilt locally.


jasikpark:
Superseded by #407, which moves the pin straight to nebula master `7cfa47d`.

